PR #57, opened 2019-09-14
Display number of videos in queue

### LINKED ISSUE #22: Display number of videos queued
This should be displayed in a bubble next to the "Queue" text

carpenterd777:
A bubble was added within the "Queue" tab that tracks the number of videos currently in the queue. A queue with over 99 videos will display "99+" instead.

closes #22

### DIFF
--- a/src/views/Room.vue
+++ b/src/views/Room.vue
@@ -31,7 +31,7 @@
             <v-tabs grow v-model="queueTab">
               <v-tab>
                 Queue
-                <span class="bubble">{{$store.state.room.queue.length < 99? $store.state.room.queue.length : "99+"}}</span>
+                <span class="bubble">{{ $store.state.room.queue.length <= 99 ? $store.state.room.queue.length : "99+" }}</span>
               </v-tab>
               <v-tab>Add</v-tab>
             </v-tabs>
@@ -98,9 +98,7 @@ export default {
       inputAddUrlText: "",
 
       showJoinFailOverlay: false,
-      joinFailReason: "",
-
-      totalItemsInQueue: 0
+      joinFailReason: ""
     };
   },
   computed: {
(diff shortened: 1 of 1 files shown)


src/views/Room.vue
```diff
@@ -95,7 +98,7 @@ export default {
       inputAddUrlText: "",
 
       showJoinFailOverlay: false,
-      joinFailReason: ""
+      joinFailReason: "",
```
dyc3: For future reference, avoid making changes like this that don't pertain to the feature/bug you are working on. It helps to avoid merge conflicts and keep commits clean.

src/views/Room.vue
```diff
@@ -265,4 +268,17 @@ export default {
   width: 100%;
   height: 100%;
 }
+.bubble{
+  height: 25px;
+  width: 25px;
+  left: 10px;
```
dyc3: I do like the styling you've done for the bubble here. Adding a little bit of margin on the left side of the bubble would make it even better.

Since `left` doesn't do anything here, just replace it with margin-left.

src/views/Room.vue
```diff
@@ -29,7 +29,10 @@
         <v-layout row justify-space-between>
           <v-flex column md8 sm12>
             <v-tabs grow v-model="queueTab">
-              <v-tab>Queue</v-tab>
+              <v-tab>
+                Queue
+                <span class="bubble">{{$store.state.room.queue.length < 99? $store.state.room.queue.length : "99+"}}</span>
```
dyc3: Good job on finding where the room state is. Just a couple of nitpicks:

1. If there are are 99 videos in the room, then it shows "99+". In this situation, I think it should say "99" instead. Change `< 99` to `<= 99` 
2. The spacing is a little bit inconsistent with the rest of the code. There should be a space after the `{{` and before the `}}`, and both sides of the `?` should have a space.

carpenterd777:
Thanks for the advice, I'll keep it in mind as I'm working on other features in the future. How do you amend a previous commit with new changes?

carpenterd777:
I attempted to amend my previous commit using `git commit --amend` as you suggested, and made all of the changes you requested to the code for neatness.

PR #60, opened 2019-09-15
youtube player immediately begins loading fixes #38

### LINKED ISSUE #38: Videos skip the first 1 or 2 seconds when a video starts playing
This is caused by the clients not having the video buffered when playback starts.

When a user hits play, the server starts incrementing `playbackProgress` immediately. When the clients receive the sync message to start playing, they start buffering the video. This usually means at least a second for the client to actually start playing the video. As a result, the user sees the video skip.

dyc3:


dyc3:
That's a good point @adapap However, with the current way the server is set up, one client with a poor connection will not affect other clients because `playbackPosition` is incremented by the server. That being said, we don't have any way to test a poor connection scenario. It's a little out of scope for this PR though.

PR #64, opened 2019-09-16
queue endpoint supports service and id parameters, closes #63

### LINKED ISSUE #63: POST /room/:name/queue should accept args service and id
Currently, this endpoint only accepts `url`. This endpoint should also support adding videos to the queue by specifying `service` and `id`, where `service` refers to what service the video is hosted on, and `id` is the id of the video on that service.

dyc3:


api.js
```diff
@@ -99,10 +99,25 @@ module.exports = function(_roommanager, storage) {
 			});
 			return;
 		}
-		let success = roommanager.addToQueue(req.params.name, req.body.url);
-		res.json({
-			success
-		});
+		if (req.body.url) {
```
adapap: Is there any reason why you can't create a response which just leaves null values (empty string in JSON) for parameters that are not provided? I don't think it's a good idea to mix data types when it comes to API endpoints.

adapap: Is there any reason why you can't create a response which just leaves null values (empty string in JSON) for parameters that are not provided? I don't think it's a good idea to mix data types when it comes to API endpoints.

roommanager.js
```diff
@@ -132,11 +132,20 @@ module.exports = function (server, storage) {
 			length: 0
 		};
 
-		queueItem.service = InfoExtract.getService(link);
+		if (typeof video === "string") {
```
adapap: Following the above comment, inconsistency in the data leads to these hacky solutions. This should check if a url is provided in the object and redirecting to other handlers if it is not.

dyc3: That's a good point, I agree. I'll rework it

dyc3:
In addition to the changes requested, I also fixed an issue with the responses for this endpoint. It would return `{"success":{}}` instead of `{"success":true}`

api.js
```diff
@@ -99,10 +99,27 @@ module.exports = function(_roommanager, storage) {
...
+			roommanager.addToQueue(req.params.name, { url: req.body.url }).then(success => {
+				res.json({
```
ifkb99: Is there a reason we are returning a json object here instead of a 200 response? If the client has logic that relies on it I am not sure it would be worth changing,

dyc3: The default status code for responses is 200. We return a json object just to return _something_. It makes it easier to debug if we are making manual requests with curl or postman.

PR #66, opened 2019-09-16
Added timestamp for current video playing

carpenterd777:
Added a timestamp for the current video to the left of the volume slider. It changes from a mm:ss to a hh:mm:ss format depending on how long the video is and the current position of the slider.

### DIFF
--- a/src/timestamp.js
+++ b/src/timestamp.js
@@ -1,0 +1,5 @@
+
+export default function convertSecondstoHHMMSS(seconds){    //if the time passed to it is less than an hour, it converts to mm:ss instead
+    let timeString = new Date(seconds * 1000).toISOString();
+    return seconds >= 3600 ? timeString.substr(11,8) : timeString.substr(14,5);
+}
--- a/src/timestamp.js
+++ b/src/timestamp.js
@@ -1,5 +1,5 @@
 
 export default function secondsToTimestamp(seconds) {    //formats seconds into mm:ss if less than an hour, hh:mm:ss if greater than an hour
-    let timeString = new Date(seconds * 1000).toISOString();
+    const timeString = new Date(seconds * 1000).toISOString();
     return seconds >= 3600 ? timeString.substr(11,8) : timeString.substr(14,5);
 }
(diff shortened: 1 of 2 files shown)


carpenterd777:
Okay, so I moved the logic to a different module, cleaned up the needlessly long lines, and changed the name of the property to `timestampDisplay`. How does it look to you now?

Also, I noticed that the time that is displayed when the user is seeking is also formatted in total seconds as opposed to a mm:ss format. I don't see it on the board, do you want me to fix it before this PR is merged or start a new PR for that feature?

src/views/Room.vue
```diff
@@ -80,6 +83,7 @@
 <script>
 import { API } from "@/common-http.js";
 import VideoQueueItem from "@/components/VideoQueueItem.vue";
+import convertSecondstoHHMMSS from "../timestamp.js";
```
dyc3: Change to `@/timestamp.js`

src/timestamp.js
```diff
@@ -0,0 +1,5 @@
+
+export default function convertSecondstoHHMMSS(seconds){    //if the time passed to it is less than an hour, it converts to mm:ss instead
```
dyc3: This function name is kinda long and kinda weird to read. Maybe `secondsToTimestamp`.

src/views/Room.vue
dyc3: `@` refers to the base directory, so if we move this file to a different directory in the future, we don't have to modify our imports.

carpenterd777:
Okay, I changed the name of the exported function and referred the base directory rather than the directory above as recommended. I also added a space before the brace.

About the slider thing I mentioned in an earlier comment, did you want me to fix that in this branch or start a different one?

dyc3:
Create a different issue for that, it's out of scope for this issue. I'd wait until this PR is merged in.

src/timestamp.js
```diff
@@ -0,0 +1,5 @@
+
+export default function secondsToTimestamp(seconds) {    //formats seconds into mm:ss if less than an hour, hh:mm:ss if greater than an hour
```
ifkb99: Looks good! The only thing I would change is to declare timeString as a const, as it is never being modified.

carpenterd777: Thanks! I'll make sure to fix that

PR #75, opened 2019-09-20
Add `production` data to Vue store, fixed #43

### LINKED ISSUE #43: Make test video buttons only appear in dev environment

adapap:
Can't figure out why the spacing got messed up because it looks normal in my editor.

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -10,7 +10,8 @@ export default new Vuex.Store({
 			message: '',
 			reconnectError: false,
 		},
-		joinFailureReason: null,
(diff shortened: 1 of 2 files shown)


adapap:
I couldn't find anywhere verifying that. Here's an article I used: https://cli.vuejs.org/guide/mode-and-env.html#using-env-variables-in-client-side-code

dyc3:
@adapap Please redo this PR, the original branch doesn't exist anymore and I don't think that it actually fixes the issue

PR #76, opened 2019-09-20
Issues with secondsToTimestamp

adapap:
Two issues caused this function to error in development:
- A zero division error when playback duration was 0 (default)
- Attribute error when accessing length property of a source that doesn't exist (default empty object)

This PR fixes both of these issues.


adapap:
Today I learned you can't delete pull requests. Have a laugh.

PR #77, opened 2019-09-20
replaced playbackDuration with currentSource.length where appropriate

### LINKED ISSUE #49: Should infer playbackDuration from currentSource.length
The current situation expects the property `playbackDuration` to be defined in each room. The current workaround is detailed in a comment here:

https://github.com/dyc3/opentogethertube/blob/b2102765bf80c982278ea408b3ee5bc1cc6bef39/roommanager.js#L40

To fix this, any reference to `playbackDuration` should be replaced with a reference to `currentSource.length`

KennethSkelton:
fixes #49

### DIFF
--- a/roommanager.js
+++ b/roommanager.js
@@ -15,7 +15,6 @@ module.exports = function (server, storage) {
 			queue: room.queue,
 			isPlaying: room.isPlaying,
 			playbackPosition: room.playbackPosition,
-			
 			users: [],
 		};
 
@@ -35,9 +34,6 @@ module.exports = function (server, storage) {
 	}
 
 	function updateRoom(room) {
-		
-			
-
 		if (_.isEmpty(room.currentSource) && room.queue.length > 0) {
 			room.currentSource = room.queue.shift();
 		}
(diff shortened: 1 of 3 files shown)


roommanager.js
```diff
@@ -35,24 +35,20 @@ module.exports = function (server, storage) {
...
-		else if (!_.isEmpty(room.currentSource) && room.playbackPosition > room.playbackDuration) {
+		else if (!_.isEmpty(room.currentSource) && room.playbackPosition > room.currentSource.length) {
 			room.currentSource = room.queue.length > 0 ? room.queue.shift() : {};
 			room.playbackPosition = 0;
 		}
 		if (_.isEmpty(room.currentSource) && room.queue.length == 0 && room.isPlaying) {
 			room.isPlaying = false;
 			room.playbackPosition = 0;
-			room.playbackDuration = 0;
+			room.currentSource.length = 0;
```
dyc3: This doesn't need to be set because it will be overwritten anyway by the next video.

PR #81, opened 2019-09-20
Fix vue linting

dyc3:


dyc3:
`lint-ci` is for travis. It won't automatically fix stuff, so it lists out all the errors in the log.

PR #82, opened 2019-09-21
Added more eslint styles to enforce code styling

adapap:
Summary of changes:
- Standardized spacing/quotes in eslint config file (pure OCD)
- Camelcase rule: seems to be the naming convention, so why not enforce it
- Indent: susceptible to change/opinions, but having a consistent spacing pattern is better (currently, 4-wide tab)
- Minimize empty lines: don't need empty lines at beginning/end of files, and no more than 1 in between blocks
- Add `.vscode/settings.json` to `.gitignore`: changing settings for the workspace causes this file to appear
  - I found the fix for `vue/no-parsing-error`  as a result of misinterpreting `<=` as a Vue component. It is a setting in the Vetur extension: `"vetur.validation.template": false`


dyc3:
I agree with adding the Camelcase rule and minimize empty lines rule. However, I'm not to keen on adding any indentation rules or quote enforcement rules right now, especially with the amount of potential merge conflicts with existing pull requests, and stuff that people are working on.

Also, the `eol-last` enforces an extra new line at the end of each file. This is a unix-ism that makes it easier to concatenate files on the command line, and not something I particularly want to change.

PR #84, opened 2019-09-22
Add local cache for video metadata

### LINKED ISSUE #24: Local cache for video info
There should be some local caching going on so we can minimize requests to the youtube api. It should be handled inside `infoextractor.js`

In the youtube api, ETags are used to determine if a resource has changed.
https://developers.google.com/youtube/v3/getting-started#etags

dyc3:
The purpose of this PR is to minimize usage of the youtube API. Closes #24 

Summary:
- Video metadata cache. When video information is requested by calling `getVideoInfo()`, we first check the database to see if it's cached already. If its not there, we retrieve it.
- Partial information retrieval. If certain information is missing from the cache, we can retrieve the bare minimum information needed to fill the gaps.
- `test` environment database config for unit tests to use.
- Unit tests to test caching behavior.

### DIFF
--- a/infoextract.js
+++ b/infoextract.js
@@ -165,6 +165,12 @@ module.exports = {
 					}
 					results.push(video);
 				}
+
+				// update cache
+				for (const video of results) {
+					storage.updateVideoInfo(video);
+				}
+
 				resolve(results);
 			}).catch(err => {
 				reject(err);
--- a/infoextract.js
+++ b/infoextract.js
@@ -43,9 +43,7 @@ module.exports = {
 
 			if (video.service === "youtube") {
 				return this.getVideoInfoYoutube([video.id], missingInfo).then(result => {
-					for (let field of missingInfo) {
-						video[field] = result[video.id][field];
-					}
+					video = Object.assign(video, result[video.id]);
 					return video;
 				}).catch(err => {
 					console.error("Failed to get youtube video info:", err);
--- a/infoextract.js
+++ b/infoextract.js
@@ -13,13 +13,37 @@ module.exports = {
 	/**
 	 * Gets all necessary information needed to represent a video. Handles
 	 * local caching and obtaining missing data from external sources.
+	 * Note that this function does not update the cache. That should be
+	 * handled by the functions that actually retrieve missing info.
 	 * @param	{string} service The service that hosts the source video.
 	 * @param	{string} id The id of the video on the given service.
 	 * @return	{Object} Video object
 	 */
 	getVideoInfo(service, id) {
-		storage.getVideoInfo(service, id).then(result => {
-			return result;
+		// TODO: check if service is valid
+		// TODO: check if id is valid for service
+		return storage.getVideoInfo(service, id).then(result => {
+			let video = {};
+			for (const key in result) {
+				if (result.hasOwnProperty(key)) {
+					video[key] = result[key];
+				}
+			}
+			let missingInfo = [
+				"title",
+				"description",
+				"thumbnail",
+				"length",
+			].filter(p => !video.hasOwnProperty(p));
+			if (missingInfo.length === 0) {
+				return video;
+			}
+
+			if (video.service === "youtube") {
+				return this.getVideoInfoYoutube([video.id]).then(result => {
+					return result[video.id];
+				});
+			}
 		}).catch(err => {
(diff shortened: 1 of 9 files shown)


migrations/20190921194802-create-cached-video.js
```diff
@@ -0,0 +1,43 @@
...
+      id: {
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.INTEGER,
+      },
+      service: {
+        type: Sequelize.STRING,
+      },
+      service_id: {
```
adapap: camelCase on `service_id` here?

infoextract.js
```diff
@@ -2,13 +2,53 @@ const axios = require("axios");
...
+	getVideoInfo(service, id) {
+		// TODO: check if service is valid
+		// TODO: check if id is valid for service
+		return storage.getVideoInfo(service, id).then(result => {
+			let video = {};
+			for (const key in result) {
+				if (result.hasOwnProperty(key)) {
+					video[key] = result[key];
+				}
+			}
```
adapap: What is the point of checking if `result` has a property `key`? Wouldn't this be better written as `const key in Object.keys(result)`?

Additionally, updating an object with partial properties sounds like a perfect use case for [`Object.assign`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign).

infoextract.js
```diff
@@ -2,13 +2,53 @@ const axios = require("axios");
...
+			let missingInfo = [
+				"title",
+				"description",
+				"thumbnail",
+				"length",
+			].filter(p => !video.hasOwnProperty(p));
```
adapap: Might want to find an alternative to hard-coding this if video properties get changed in the future. Maybe from existing cache data or from database model?

infoextract.js
dyc3: The `if (result.hasOwnProperty(key))` came from the vscode template. And yes, `Object.assign` appears to be the better solution.

infoextract.js
dyc3: That's a good idea, I'll look into it.

migrations/20190921194802-create-cached-video.js
dyc3: Yeah I'll change that

storage.js
```diff
@@ -33,7 +35,82 @@ module.exports = {
...
+			// less than 7 days ago, then the interval is 7 days. The reason for this is that the uploader
+			// is unlikely to change the video info after a week of the original upload. Since we don't store
+			// the upload date, we pretend the original cache date is the upload date. This is potentially an
+			// over optimization.
+			const isCachedInfoValid = lastUpdatedAt.diff(today, "days") <= (origCreatedAt.diff(today, "days") <= 7) ? 7 : 30;
+			let video = {
+				service: cachedVideo.service,
+				id: cachedVideo.serviceId,
+			};
+			// We only invalidate the title and description because those are the only ones that can change.
```
adapap: The thumbnail can also change. The video author may also unpublish a video; might need to test what the API returns in this case.

storage.js
```diff
@@ -33,7 +35,82 @@ module.exports = {
...
+			}).catch(err => {
+				console.error("Failed to cache video info", err);
+				return false;
+			});
+		});
+	},
+	getVideoInfoFields() {
+		let fields = [];
+		for (let column in CachedVideo.rawAttributes) {
+			if (column === "id" || column === "createdAt" || column === "updatedAt" || column === "serviceId") {
```
adapap: Just a side-note: for this pattern I usually do something equivalent to `[...].contains(column)`

storage.js
dyc3: The thumbnail may change, but the _url_ to the thumbnail doesn't (needs confirmation).
For example, this video: https://www.youtube.com/watch?v=mOVOS9AjgFs
Has thumbnail: https://img.youtube.com/vi/mOVOS9AjgFs/hqdefault.jpg

storage.js
dyc3: `rawAttributes` is an object

adapap: In that case you can use `Object.keys(...)` to make the pattern if you want to use the filter method.

Issue #65, opened 2019-09-16
JS for each pattern instead of looping by index


adapap:
https://github.com/dyc3/opentogethertube/blob/e15561900a576ff45adee93cd323a4fd90a7c2b1/infoextract.js#L72-L89

Javascript has two enumeration keywords that can be used in for loops: `in` and `of`
The former is used to enumerate by index for arrays or keys in objects, and the latter can be used to iterate through an object, such as values of an array.

I propose enforcing the use of [for .. of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) to increase readability/reduce redundancy in code. I am merely using the above as an example; there are many occurrences of the same construct.

```
for (let r = 0; r < regexs.length; r++) { 
    let matches = res.data.match(regexs[r]); 
    ... 
```
This can be replaced with:
```
for (let regex of regexs) {
    let matches = res.data.match(regex);
    ...
```

dyc3:
I avoided using these keywords for 2 reasons. Firstly, I found `for ... in` and `for ... of` syntax confusing and actually less readable than just typical `for` syntax. Secondly, when writing a loop, if I suddenly wanted access to the index its available readily without needing to rewrite the `for` statement.

In this particular example, you can't replace
```
for (let r = 0; r < regexs.length; r++) { 
    let matches = res.data.match(regexs[r]); 
    ... 
```
because `r` is used later on line 82.

But I'd like to hear what others think.

benwaffle:
use `for .. of` if you don't need the numerical index. don't use `for ... in` ever

adapap:
Actually I disagree, `for .. in` can be used to loop through objects by key.

benwaffle:
nah, you want `for (const key of Object.keys(obj))`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#Why_Use_for...in

ifkb99:
I think both ideas can be combined, in a way, for more readable and efficient code.

About halfway down, [this article](https://stackoverflow.blog/2019/09/12/practical-ways-to-write-better-javascript/) explains how basic for loops slow down code and increase complexity. With the forEach higher order function, you can still get the index along with completing the loop  asynchronously (this would obviously not work for sections that need to be completed sequentially).

[...]

PR #86, opened 2019-09-22
fix crash when sending sync to closed socket

dyc3:
Sometimes, the room manager doesn't remove clients right away. If a client disconnects and the server attempts to send a message before the room manager removes the client, the whole server crashes. This fixes this issue by adding a check before sync messages are sent.

### DIFF
--- a/roommanager.js
+++ b/roommanager.js
@@ -19,17 +19,22 @@ module.exports = function (server, storage) {
 			users: [],
 		};
 
-
 		for (let i = 0; i < room.clients.length; i++) {
+			let ws = room.clients[i].socket;
+
+			// make sure the socket is still open
+			if (ws.readyState != 1) {
+				continue;
+			}
+
 			syncMsg.users = [];
 			for (let u = 0; u < room.clients.length; u++) {
 				syncMsg.users.push({
 					name: room.clients[u].name,
-					isYou: room.clients[i].socket == room.clients[u].socket,
+					isYou: ws == room.clients[u].socket,
 				});
 			}
 
-			let ws = room.clients[i].socket;
 			ws.send(JSON.stringify(syncMsg));
 		}
 	}


roommanager.js
```diff
@@ -19,17 +19,22 @@ module.exports = function (server, storage) {
 			users: [],
 		};
 
-
 		for (let i = 0; i < room.clients.length; i++) {
+			let ws = room.clients[i].socket;
```
adapap: `for (let ws of room.clients) { ...`

roommanager.js
```diff
@@ -19,17 +19,22 @@ module.exports = function (server, storage) {
...
+
+			// make sure the socket is still open
+			if (ws.readyState != 1) {
+				continue;
+			}
+
 			syncMsg.users = [];
 			for (let u = 0; u < room.clients.length; u++) {
```
adapap: `for (let user of room.clients) { ...`

`i` and `u` do not need to be used and are ambiguous.

dyc3:
@adapap We never came to a conclusion on #65. Plus, it's out of scope for the purpose of this PR.

roommanager.js
```diff
@@ -19,17 +19,22 @@ module.exports = function (server, storage) {
 			users: [],
 		};
 
-
 		for (let i = 0; i < room.clients.length; i++) {
```
ifkb99: I know I'm sounding like a broken record here, but this really should be a forEach loop, especially in this section of code. The forEach function allows the loop to run asynchronously, while normal for loops block. As we are dealing with websockets here and not just arrays, this is a performance issue and not simply a styling one.

My comment also applies to the loop syncing clients down below

dyc3: The purpose of this PR in particular is to fix a bug. I'd be happy to refactor in a separate PR, but I want this fixed because production goes down quite frequently due to this bug.

PR #89, opened 2019-09-23
Added support for youtube channels

ifkb99:
Refactored some stuff by accident

### DIFF
--- a/infoextract.js
+++ b/infoextract.js
@@ -75,11 +75,12 @@ module.exports = {
 				`${channelData['channel'] ? 'id' : 'forUsername'}=${channelData["id"]}`
 				//if the link passed is a channel link, ie: /channel/$CHANNEL_ID, then the id filter must be used
 				//on the other hand, a user link requires the forUsername filter
-			).then(async res => {
+			).then(res => {
 				if (res.status === 200) {
-					resolve(await this.getPlaylistYoutube(
+					this.getPlaylistYoutube(
 						res.data.items[0].contentDetails.relatedPlaylists.uploads
-					));
+					).then(playlist => resolve(playlist))
+					.catch(err => reject(err));
 				}
 				else {
 					reject(`Failed with status code ${res.status}`);
@@ -161,8 +162,8 @@ module.exports = {
 	getVideoLengthYoutube_Fallback: async (url) => {
 		let res = await axios.get(url);
 		let regexs = [
-/length_seconds":"\d+/, /lengthSeconds\\":\\"\d+/,
-];
+			/length_seconds":"\d+/, /lengthSeconds\\":\\"\d+/,
+		];
 		for (let r = 0; r < regexs.length; r++) {
 			let matches = res.data.match(regexs[r]);
 			if (matches == null) {
@@ -219,18 +220,15 @@ module.exports = {
 		});
 	},
 
-	async getManyPreviews(videoIds) {
-		try {
-			const videoInfo = await this.getVideoInfoYoutube(videoIds);
-			return videoIds.map(id => videoInfo[id])
-						.filter(info => info !== undefined); //remove deleted videos
-		}
-		catch (err) {
-			console.error("Failed to compile add preview: error getting video info:", err);
-		}
+	getManyPreviews(playlist) {
+		//get info for all videos in playlist, then return all non-deleted videos
+		return Promise.all(
+			playlist.map(video => video.id).map(id => this.getVideoInfo('youtube', id))
+		).then(previews => previews.filter(item => item !== undefined))
+		.catch(err => console.error('Error getting video info:', err));
 	},
 
-	async getAddPreview(input) {
+	getAddPreview(input) {
 		const service = this.getService(input);
 
 		if (service !== "youtube") {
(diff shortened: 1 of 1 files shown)


infoextract.js
```diff
@@ -124,51 +144,56 @@ module.exports = {
...
 				}).catch(err => {
 					console.error("Failed to compile add preview: error getting playlist:", err);
 					reject(err);
 				});
 			});
-		}
-		else {
+		} else if (urlParsed.path.startsWith('/user') || urlParsed.path.startsWith('/channel')) {
+			console.log('channel found');
+			return new Promise((resolve, reject) => {
```
dyc3: There is no need to create another promise here, you can just return the promise from `getChannelInfoYoutube`

infoextract.js
```diff
@@ -124,51 +144,56 @@ module.exports = {
...
 
+	getManyPreviews(videoIds, qParamsV) {
+		return new Promise(async (resolve, reject) => {
+			try {
+				const videoInfo = await this.getVideoInfoYoutube(videoIds);
+				resolve(videoIds.map(id => videoInfo[id])
+						   	.filter(info => info !== undefined) //remove deleted videos
+						   	.map(info => {
+									if (qParamsV && qParamsV === info.id) {
+										info.highlight = true;
```
dyc3: I don't think `highlight` should be set in here, it should be done in `getAddPreview`

infoextract.js
dyc3: There are several other instances of this used in your PR, please fix those as well.

infoextract.js
```diff
@@ -29,6 +29,26 @@ module.exports = {
 		}
 	},
 
+	getChanneInfoYoutube(link) {
```
dyc3: Don't pass in the full link here, instead pass in the channel ID.

infoextract.js
```diff
@@ -124,51 +144,56 @@ module.exports = {
 		});
 	},
 
+	getManyPreviews(videoIds, qParamsV) {
```
dyc3: I like this.

infoextract.js
ifkb99: Why is that? I did it here so that it will run whenever the function is called, as it needs the video info to work correctly

infoextract.js
ifkb99: I didn't pass in just the channel id in order to support links to users as well. If you look a few lines down you can see this when I parse the link to find if it starts with 'channel' or 'users'

infoextract.js
dyc3: It makes the function less usable in other places, if we wanted to do that, because you need to pass in the query parameters from the url.

infoextract.js
dyc3: Pass in an object with either `channel` or `user` set.

infoextract.js
dyc3: It might be better to remove `highlight` altogether for now because it's currently not used.

infoextract.js
ifkb99: That sounds like a better idea. A channel url looks like .../channel/${channel_id}, where a user url looks like .../user/${youtube_username}

dyc3:
@ifkb99 Please rebase this branch on the current master and make the appropriate changes.

dyc3:
@ifkb99 It looks like the tests are failing because the environment variable `YOUTUBE_API_KEY` is undefined. This is probably because this PR comes from your fork, so to prevent leaking the api key travis just doesn't set it.

Since tests are passing locally, we'll just ignore the failed travis check when we merge. Please fix the issues with creating extra promises as mentioned above, and we can merge it.

infoextract.js
```diff
@@ -196,50 +220,56 @@ module.exports = {
...
 				}).catch(err => {
 					console.error("Failed to compile add preview: error getting playlist:", err);
 					reject(err);
 				});
 			});
 		}
+		else if (urlParsed.path.startsWith('/user') || urlParsed.path.startsWith('/channel')) {
+			console.log('channel found');
+			try {
+				const channelData = {id: urlParsed.path.slice(urlParsed.path.lastIndexOf('/')+1)};
```
dyc3: I can see what you did here and why, but it's a little inconsistent with how we handle things like this in other parts of the project. The way I would have done it is just use the properties `channel` or `user` to indicate what the value is. Something like:
```
const channelId = urlParsed.path.slice(urlParsed.path.lastIndexOf('/') + 1);
if (urlParsed.path.startsWith('/channel/')) {
	channelData.channel = channelId;
}
else {
	channelData.user = channelId;
}
```

Also, add spaces around `+` and the inside of `{}` here

infoextract.js
```diff
@@ -196,50 +220,56 @@ module.exports = {
...
+				if (urlParsed.path.startsWith('/channel/')) {
+					channelData.channel = true;
+				}
+				else {
+					channelData.user = true;
+				}
+				return this.getChanneInfoYoutube(channelData)
+					.then(newestVideos => this.getManyPreviews(newestVideos))
+					.then(previews => previews)
```
dyc3: You aren't doing anything with this `then`, I'm pretty sure you can get rid of it.

infoextract.js
```diff
@@ -67,6 +67,30 @@ module.exports = {
 		}
 	},
 
+	getChanneInfoYoutube(channelData) {
+		return new Promise((resolve, reject) => {
```
dyc3: This is an extra promise, just `return` instead of `resolve`. Promises chain, so if you return in one of the `then` or `catch` callbacks, the returned value works its way up the chain.

infoextract.js
```diff
@@ -196,50 +220,56 @@ module.exports = {
 		});
 	},
 
+	getManyPreviews(playlist) {
+		//get info for all videos in playlist, then return all non-deleted videos
+		return Promise.all(
+			playlist.map(video => video.id).map(id => this.getVideoInfo('youtube', id))
```
dyc3: You could do this in one `map`

PR #103, opened 2019-09-28
Add lots of tests for back end

### LINKED ISSUE #13: Add unit tests

dyc3:
I added enough tests to feel comfortable enough to close #13


infoextract.js
```diff
@@ -48,7 +47,15 @@ module.exports = {
...
 
 	getService(link) {
+		if (typeof link !== "string") {
+			return false;
+		}
+
 		let srcUrl = url.parse(link);
+		if (srcUrl.host === null) {
+			return false;
+		}
```
adapap: Is this temporary until we add more services? If not, I would make sure that either the function returns empty strings instead of false, or raises an error.

tests/unit/infoextractor.spec.js
```diff
@@ -9,7 +10,47 @@ if (!fs.existsSync(config_path)) {
...
+    expect(InfoExtract.getService("https://www.youtube.com/watch?v=I3O9J02G67I")).toEqual("youtube");
+    expect(InfoExtract.getService("https://m.youtube.com/watch?v=I3O9J02G67I")).toEqual("youtube");
+    expect(InfoExtract.getService("http://youtu.be/I3O9J02G67I")).toEqual("youtube");
+    expect(InfoExtract.getService("https://youtu.be/I3O9J02G67I")).toEqual("youtube");
+  });
+
+  it('getService() should return false when given link to unsupported service', () => {
+    expect(InfoExtract.getService("http://example.com")).toEqual(false);
+  });
+
```
adapap: Refer to previous comment

tests/unit/roommanager.spec.js
```diff
@@ -0,0 +1,129 @@
+const fs = require('fs');
+const path = require('path');
+const roommanager = require("../../roommanager")({ on() {} }, require("../../storage"));
```
adapap: What does this do?

tests/unit/roommanager.spec.js
```diff
@@ -0,0 +1,129 @@
...
+const InfoExtract = require("../../infoextract");
```
adapap: Aside: This is imported as PascalCase whereas roommanager is all lowercase?

tests/unit/roommanager.spec.js
```diff
@@ -0,0 +1,129 @@
...
+
+const config_path = path.resolve(process.cwd(), `env/${process.env.NODE_ENV}.env`);
```
adapap: camelCase lint rule, just run linter probably.

tests/unit/roommanager.spec.js
```diff
@@ -0,0 +1,129 @@
...
+    });
+  });
+
+  it('should add a video to the queue with service and id provided, and because a video is playing, leave it in the queue', done => {
+    InfoExtract.getVideoInfo = jest.fn().mockReturnValue(new Promise(resolve => resolve({ service: "youtube", id: "I3O9J02G67I", length: 10 })));
+    expect(roommanager.rooms).toBeDefined();
+    expect(roommanager.rooms.test).toBeDefined();
+    expect(roommanager.rooms.test.name).toBeDefined();
+    expect(roommanager.rooms.test.name.length).toBeGreaterThan(0);
+    expect(roommanager.rooms.test.queue.length).toEqual(0);
```
adapap: I don't usually write tests but is there a way to reduce this repetition?

infoextract.js
dyc3: It's to ensure that `link` is a valid URL

tests/unit/roommanager.spec.js
dyc3: Makes it so that we can import the room manager and run it's functions. It's part of the reason of why I'm working on rewriting the room manager now.

tests/unit/roommanager.spec.js
dyc3: It's consistent with how these modules are imported in other files

tests/unit/roommanager.spec.js
dyc3: Maybe. Since I'm rewriting the room manager, this will probably get changed anyway.

PR #83, opened 2019-09-22
Updating room properties

adapap:
This PR can serve as a starting point for updating room properties (#51). Feel free to expand on this concept.
- Update room title/description from editable fields in room
- Change `room` to be a computed property which returns `this.$store.state.room`

PR #112, opened 2019-09-30
Add tests for room view and code coverage report

dyc3:


tests/unit/room.spec.js
```diff
@@ -0,0 +1,155 @@
...
+        $disconnect: jest.fn(),
+      },
+      stubs: ['youtube'],
+    });
+  });
+
+  it('should render room title in permanent rooms, if the room has one', () => {
+    store.state.room.title = 'Example Room';
+    store.state.room.isTemporary = false;
+    const roomTitle = wrapper.find('h1');
```
adapap: This is not `h1` in my PR for editable room names. Maybe consider #83 a blocker.

adapap: This is not `h1` in my PR for editable room names. Maybe consider #83 a blocker.

tests/unit/room.spec.js
```diff
@@ -0,0 +1,155 @@
...
+    expect(testVideoButtons.length).toEqual(2);
+    expect(testVideoButtons.at(0).text()).toEqual('Add test video 0');
+    expect(testVideoButtons.at(1).text()).toEqual('Add test video 1');
+  });
+
+  it('should NOT render test buttons when in production environment', () => {
+    store.state.production = true;
+    const testVideoButtons = wrapper.find('.video-add').findAll({ name: 'v-btn' });
+    expect(testVideoButtons.length).toEqual(0);
+  });
```
adapap: Possibly merge these two tests into one?

src/views/Room.vue
```diff
@@ -22,7 +22,7 @@
                   <v-icon>fas fa-fast-forward</v-icon>
                 </v-btn>
                 <vue-slider v-model="volume" style="width: 150px; margin-left: 10px"/>
-                <div style="margin-left: 20px">
+                <div style="margin-left: 20px" class="timestamp">
```
adapap: Generally, if you have a class for an element you don't want to also have inline styles.

dyc3: I think that instances like this, where I'm setting something like `margin-left` its easier to debug later, and the class can be used in other places.

tests/unit/room.spec.js
dyc3: #83 is a draft. Just rebase your PR onto master when this is merged and update the test so that it works.

tests/unit/room.spec.js
dyc3: These tests verify 2 distinct behaviors

PR #125, opened 2019-10-20
make onInputAddChange() run without needing to defocus the text box

### LINKED ISSUE #41: Add preview only appears after the user defocuses the text box

dyc3:
Changed the way that `onInputAddChange()` was triggered. Instead of using the `@change` event, we use a watcher on `inputAddUrlText` to trigger it.

Fixes #41

### DIFF
--- a/src/views/Room.vue
+++ b/src/views/Room.vue
@@ -46,7 +46,7 @@
               </v-tab-item>
               <v-tab-item>
                 <div class="video-add">
-                  <v-text-field placeholder="Video URL to add to queue" @change="onInputAddChange" v-model="inputAddUrlText"/>
+                  <v-text-field placeholder="Video URL to add to queue" v-model="inputAddUrlText"/>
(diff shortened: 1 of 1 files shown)


src/views/Room.vue
```diff
@@ -300,6 +307,11 @@ export default {
         this.$refs.youtube.player.seekTo(newPosition);
       }
     },
+    inputAddUrlText() {
+      // HACK: The @change event only triggers when the text field is defocused.
+      // This ensures that onInputAddChange() runs everytime the text field's value changes.
+      this.onInputAddChange();
+    },
```
adapap: See @input event on `v-text-fields`

adapap: Actually, @keydown might be what I was thinking of.

dyc3: I don't think `@keydown` would be triggered if the user pastes text using the right click context menu. This method ensures we get all changes to the text field, regardless to the source of the change.

PR #131, opened 2019-10-25
secondsToTimestamp handles negative values, unit test added , babel.c…

### LINKED ISSUE #110: secondsToTimestamp does not handle negative values properly

carpenterd777:
…onfig.js edited

Fixes #110

### DIFF
--- a/src/timestamp.js
+++ b/src/timestamp.js
@@ -1,4 +1,6 @@
 export default function secondsToTimestamp(seconds) {    //formats seconds into mm:ss if less than an hour, hh:mm:ss if greater than an hour
-    const timeString = new Date(seconds * 1000).toISOString();
-    return seconds >= 3600 ? timeString.substr(11, 8) : timeString.substr(14, 5);
+    const posSeconds = Math.abs(seconds);
+    const timeString = new Date(posSeconds * 1000).toISOString();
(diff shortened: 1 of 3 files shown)


tests/unit/secondsToTimestamp.spec.js
```diff
@@ -0,0 +1,15 @@
+import secondsToTimestamp from '../../src/timestamp';
+
+describe('secondsToTimestamp spec', () => {
+    it('handles positive values', () => {
+        let s = 120;
```
dyc3: You don't have to define a new variable, you can just call the function with this value since it's only used once.

src/timestamp.js
```diff
@@ -1,4 +1,6 @@
 export default function secondsToTimestamp(seconds) {    //formats seconds into mm:ss if less than an hour, hh:mm:ss if greater than an hour
-    const timeString = new Date(seconds * 1000).toISOString();
-    return seconds >= 3600 ? timeString.substr(11, 8) : timeString.substr(14, 5);
+    const posSeconds = Math.abs(seconds);
+    const timeString = new Date(posSeconds * 1000).toISOString();
+    const subTimeString = posSeconds >= 3600 ? timeString.substr(11, 8) : timeString.substr(14, 5);
+    return seconds < 0 ? "-" + subTimeString : subTimeString;
```
dyc3: This is a little repetitive, maybe just `(seconds < 0 ? "-" : "") + subTimeString`

PR #151, opened 2019-11-17
Disable seeker slider when no video is loaded.

### LINKED ISSUE #118: Rooms with no video loaded are seekable to 1:40
I don't think this is a huge deal as far as functionality goes, just some behavior that I noticed which I'm guessing isn't desirable.

DennisvanderWerf:
Fixes #118


dyc3:
Not sure why travis isn't running, but the tests work so I'm merging it.

PR #155, opened 2019-12-06
Uses 'default' method of retrieval for private vid

### LINKED ISSUE #147: If a link to a youtube video in a private playlist is used to make an add preview, it fails without showing the video
example: https://www.youtube.com/watch?v=RymPLuA_l6A&list=PLhNByNs9KvL9edGssl21N99JVG4gLwx0E&index=2&t=0s

carpenterd777:
Intended to fix #147

### DIFF
--- a/infoextract.js
+++ b/infoextract.js
@@ -251,7 +251,14 @@ module.exports = {
 					this.getManyPreviews(playlist).then(previews => resolve(previews));
 				}).catch(err => {
 					console.error("Failed to compile add preview: error getting playlist:", err);
-					reject(err);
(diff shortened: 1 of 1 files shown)


infoextract.js
```diff
@@ -251,7 +251,14 @@ module.exports = {
 					this.getManyPreviews(playlist).then(previews => resolve(previews));
 				}).catch(err => {
 					console.error("Failed to compile add preview: error getting playlist:", err);
```
dyc3: We should change the logging here a little so that we only see stack traces if there is actually a problem

PR #282, opened 2020-08-07
Extract video player into its own component

Butt4cak3:
This pull request creates a new component that I named "OmniPlayer" (let me know if you can come up with a better name) that abstracts away the different video services that OTT supports.

The Room component currently deals with differentiating between video services, which leads to a bunch of duplicated code.

**Example:**
Before
```js
     play() { 
       if (this.currentSource.service == "youtube") { 
         this.$refs.youtube.play(); 
       } 
       else if (this.currentSource.service === "vimeo") { 
         this.$refs.vimeo.play(); 
       } 
       else if (this.currentSource.service === "dailymotion") { 
         this.$refs.dailymotion.play(); 
       } 
       else if (this.currentSource.service === "googledrive") { 
         this.$refs.googledrive.play(); 
       } 
       else if (this.currentSource.service === "direct") { 
         this.$refs.direct.play(); 
       } 
     },
```

After
```js
    play() {
      this.$refs.player.play();
    },
```

### DIFF
--- a/src/components/DailymotionPlayer.vue
+++ b/src/components/DailymotionPlayer.vue
@@ -87,7 +87,5 @@ export default {
 <style lang="scss" scoped>
 .dailymotion {
 	color: #696969;
-	border: 1px solid #666;
-	border-radius: 3px;
 }
 </style>
--- a/src/components/OmniPlayer.vue
+++ b/src/components/OmniPlayer.vue
@@ -1,0 +1,115 @@
+<template>
+  <fragment>
+    <YoutubePlayer
+      v-if="source.service == 'youtube'"
+      ref="youtube"
+      v-bind="$attrs"
+      :video-id="source.id"
+      @playing="$emit('playing')"
+      @paused="$emit('paused')"
+      @ready="$emit('ready')"
+      @buffering="$emit('buffering')"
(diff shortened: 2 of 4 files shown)


Butt4cak3:
I forgot to add that I could only kind of test this one. When I queue a video and start to play it in my local instance (npm run dev-windows), the video plays for a second or two before the player goes back to "No video is playing.". That happens regardless of whether I am on my branch or a clean master branch.

dyc3:
`roommanager` heavily relies on knowing the duration of the video in order to know when to switch to the next video. If you don't have a youtube API key, then it can't use the youtube api. It may fail to web scrape the video length from the page. Open `db/dev.sqlite` and check the `CachedVideos` table to see if `length` is greater than 0.

dyc3:
The dailymotion player isn't filling the space vertically. I believe there is some styles that need to be moved over to this component.

Butt4cak3:
Yup. There are definitely a couple of things missing. I think I broke fullscreen as well.

Butt4cak3:
I managed to fix the dailymotion player height.

While doing so, I found that previously, some of the players had borders and others did not. It was pretty much 50/50 so I went with what I personally prefer (no borders) for now.

PR #284, opened 2020-08-08
Service adapters

### LINKED ISSUE #202: InfoExtractor needs standard extractor interface
Having a common extractor interface will make it easier to add new extractors in the future. Similar to how youtube-dl implements it's extractors.

Butt4cak3:
This pull request is in a **very** early stage of development! It exists to discuss the idea before I continue to work on it.

## What is this?

Service adapters are an attempt at splitting the code in infoextract.js into smaller service files. The goal is to make infoextract.js fully agnostic in regards to video services. There should be no mention of "youtube", "dailymotion" or any other service name left.

## Why?

I don't believe that the current solution scales very well. Splitting every video service into its own module enables us to more easily add or remove services down the line. It should also result in a more consistent way of handling services. Once the system is in place, a new video service can be added by creating a file in server/services/, implementing a couple of adapter methods that query an API and telling the info extractor that the adapter exists.

## How does it work?

Every video service gets its own class that extends the `ServiceAdapter` class. Each adapter must implement a set of methods that enable communication with its respective API. Such methods include `getVideoId`, `getVideoInfo` and `getAddPreview`.

With these adapters in place, the info extractor itself can be reduced to cache management, picking the right service adapter and delegating data fetching to it.

## Some technical things

The service adapters themselves should only be concerned with fetching data from some sort of API. They don't care about cache at all. The info extractor can, however, access the cache. This way, it can either return cached results or delegate on a cache miss and cache whatever the adapter returned.

This is basically what I imagine the `getVideoInfo` function in infoextract.js to look like:

```js
async function getVideoInfo(service, id) {
  // Check cache
  let { video, missingInfo } = await getCachedVideo(service, id);
  if (missingInfo.length === 0) {
    return video;
  }

  // Cache miss, query API instead
  let adapter = getServiceAdapter(service);
  video = Video.merge(video, await adapter.getVideoInfo(id, missingInfo));
  updateCache(video);
  return video;
}
```

## Progress

- [ ] InfoExtractor
- [ ] Adapters
  - [X] YouTube
    - [X] Direct links
    - [X] Playlists
    - [X] Channels
    - [X] Video URLs with list parameters
    - [X] Search
  - [X] Vimeo
  - [X] Dailymotion
  - [ ] Google Drive
    - [ ] Direct links
    - [ ] Folders
  - [ ] Direct

Closes #202


dyc3:
I've been meaning to do something like this for a while, just wasn't exactly sure of the approach I wanted to take.

Butt4cak3:
It's difficult to get things like this right the first time. I must admit that I worked on this for about 8 hours already just to throw most of it in the trash and start over. That's why, on my second attempt, I decided to publish early to get another opinion. Either way, this is going to require *major* refactoring work and a lot of time. But I'm dedicated. I've been meaning to work on something like this since TogetherTube shut down. I just didn't know OTT existed or I would've been on-board a while ago.

server/serviceadapter.js
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    return this.constructor.SERVICE_ID;
+  }
+
+  /**
+   * A boolean that indicates whether video metadata can be safely cached.
+   * @returns {boolean}
+   */
+  get cacheSafe() {
```
dyc3, line 17:
```suggestion
  get isCacheSafe() {
```

server/serviceadapter.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  /**
+   * Fetches video metadata from the API.
+   * @param {string} videoId
+   * @returns {Promise}
+   */
+  getVideoInfo(videoId, onlyProperties) {
+    return this._getVideoInfo(videoId, onlyProperties);
+  }
```
dyc3, line 37: I don't really like having `getVideoInfo` and then `_getVideoInfo` for the derived service adapters. Can we just have `getVideoInfo`?

Butt4cak3: I don't like it either tbh. I did this to be able to utilize code completion and in-editor documentation. If `ServiceAdapter` doesn't have any reference to `getVideoInfo`, the editor won't know the method exists. If you do something like
```js
  getVideoInfo() {}
```
so that child classes can override it, the editor won't know about its parameters. The `_` version is the only way I can think of that preserves all this information.

server/serviceadapter.js
Butt4cak3: I wasn't really sure about this one. I went from `enableCache` to `cacheEnabled` to `cacheSafe` until I decided to leave it as-is and wait for input. I agree that `isCacheSafe` is better, though.

dyc3:
I think the hardest bit is going to be abstracting out the parts that grab multiple videos at a time (youtube playlists/channels, google drive folders, etc).

Butt4cak3:
> I think the hardest bit is going to be abstracting out the parts that grab multiple videos at a time (youtube playlists/channels, google drive folders, etc).

Those aren't too difficult to handle. We only have to deal with video collections in the `getAddPreview` type functions and when the user directly inputs a URL. In those cases, we can just pass the entire URL to the adapter and let it deal with whatever it finds. It can then either return a single video or a list. A list is always valid because we're dealing with the "Add" tab under the video.

The one thing I couldn't figure out so far is dealing with YouTube video links that also include a playlist id (i.e. https://youtube.com/watch?v=...&list=...). Those are direct links to single videos *and* collections at the same time. I would say that the best solution is to make the adapter treat it as a collection, but move the directly linked video to the top of the list. This way, we can present the entire playlist to the user, but the linked video is the first item in the list.

To make that work, we'll probably need an adapter method like `isLinkSingleVideo` that, when it returns `true`, triggers a cache search first. If it's false, the URL will just be passed to whatever method deals with URLs. Caching collections doesn't really make sense because they can change too easily. And even *if* the linked video is already cached, we have to hit the API anyway to get the rest of the list.

dyc3:
> The one thing I couldn't figure out so far is dealing with YouTube video links that also include a playlist id

The way we handled it before is to get the playlist, and "highlight" the video with the matching id at the top of the add preview on the client side by setting `video.highlight = true`.

dyc3:
Something that I would like to see in this implementation is to make it easier to do pagination

Butt4cak3:
I believe that I found a pretty solid way to handle the search box. Here is some pseudo-ish code:

```js
async function getSearchResults(input) {
  // input contains whatever the user typed into the search bar.
  let results = [];

  if (isURL(input)) {
    // If the input is a URL, we can determine which adapter can deal with it.
    let adapter = getServiceAdapterForURL(input);

    // Caching collections doesn't work, so we only look for single videos.
    if (!adapter.isCollectionURL(input)) {
      let cachedResult = await getCachedVideo(
        adapter.serviceId,
        adapter.getVideoId(input)
      );

      if (cachedResult) {
        results = [cachedResult];
      }
    }

    // If the results array is still empty, it means that we're either dealing with a collection
    // or with a cache miss.
    if (results.length === 0) {
      results = await adapter.getVideoInfo(input);
    }
  }
  else {
    // Everything that isn't a URL triggers a search.
    results = await performSearch(input);
  }

  // Cache what we have and return it.
  updateCache(results);
  return results;
}

function performSearch(input) {
  // Call the search method on *every* service adapter. Services that don't support searches can
  // return empty arrays.
  return Promise.all(
    adapters.map((adapter) => adapter.search(input))
  ).then((results) => results.flat());
}
```

Butt4cak3:
> Something that I would like to see in this implementation is to make it easier to do pagination

That probably means that what the code above does wouldn't work - or at least not as trivially as I thought. We'd probably need a selector switch for the service the user wants to search instead of searching all services at once.

Now that I think about it, that's probably better either way. I don't think we want to hit every API every single time a user searches for something.

Butt4cak3:
Alright. I got everything to a point where it "works". The YouTube adapter is still very incomplete because it is the most complex one of the bunch and I still can't query the YouTube API from my local testing environment. Vimeo and Dailymotion worked fine in testing for me. Direct still needs some more testing.

I generated a key using the Google API Console, but there must be something that I'm missing or doing wrong. I can't stop the API from giving me 403's. It's definitely not that I'm just out of quota because I used it too much. Even the first request of the day returns a 403. That makes it really hard to test. But direct links work, at least!

The PR is updated to include the current progress.

common/video.js
```diff
@@ -27,9 +27,6 @@ class Video {
 			delete this.description;
 			delete this.url;
 		}
-		else if (["direct"].includes(this.service)) {
```
Butt4cak3, line 30: I changed a few things here and there to make things work and I probably broke *something*. This is one of those places.

I don't understand why you deleted these properties instead of just ignoring them. I built the direct video adapter so that it stores the URL in the ID property through the `getVideoId` method. That didn't work, however, because the Video class would remove the `id` property and the URL would get lost. That's why I removed this.

server/serviceadapter.js
```diff
@@ -27,13 +27,61 @@ class ServiceAdapter {
     return false;
   }
 
+  /**
+   * Determines whether a given URL points to a collection of videos.
+   * @param {string} url
+   * @returns {boolean}
+   */
+  isCollectionURL() {
+    throw `Service ${this.serviceId} does not implement method isCollectionURL`;
```
dyc3, line 36: These should throw some kind of exception instead of just a string

server/infoextractor.js
```diff
@@ -0,0 +1,181 @@
...
+  try {
+    const result = await storage.getVideoInfo(service, videoId);
+    const video = new Video(result);
+    const missingInfo = storage
+      .getVideoInfoFields(video.service)
+      .filter(p => !video.hasOwnProperty(p));
+
+    if (video.mime && !this.isSupportedMimeType(video.mime)) {
+      throw new UnsupportedMimeTypeException(video.mime);
+    }
```
dyc3, line 45: Definitely not the right place to throw this exception

dyc3:
> I still can't query the YouTube API from my local testing environment.

Check to see if `env/development.env` has the API key like this

```
YOUTUBE_API_KEY=AI32km....
```

Other than that, I'm not sure what the problem could be.

server/services/dailymotion.js
```diff
@@ -0,0 +1,54 @@
+const URL = require("url");
+const axios = require("axios");
+const ServiceAdapter = require("../serviceadapter");
+const { InvalidVideoIdException } = require("../exceptions");
+const Video = require("../../common/video");
+
+class DailyMotionAdapter extends ServiceAdapter {
+  static SERVICE_ID = "dailymotion";
```
dyc3, line 8: I'm getting a syntax error here for some reason

dyc3:
```
/home/carson/Documents/code/opentogethertube/server/services/dailymotion.js:8
  static SERVICE_ID = "dailymotion";
                    ^

SyntaxError: Unexpected token =
    at new Script (vm.js:80:7)
    at createScript (vm.js:274:10)
    at Object.runInThisContext (vm.js:326:10)
    at Module._compile (internal/modules/cjs/loader.js:664:28)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:712:10)
    at Module.load (internal/modules/cjs/loader.js:600:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:539:12)
    at Function.Module._load (internal/modules/cjs/loader.js:531:3)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.<anonymous> (/home/carson/Documents/code/opentogethertube/server/infoextractor.js:3:28)
    at Module._compile (internal/modules/cjs/loader.js:701:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:712:10)
    at Module.load (internal/modules/cjs/loader.js:600:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:539:12)
    at Function.Module._load (internal/modules/cjs/loader.js:531:3)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.<anonymous> (/home/carson/Documents/code/opentogethertube/roommanager.js:6:21)
    at Module._compile (internal/modules/cjs/loader.js:701:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:712:10)
    at Module.load (internal/modules/cjs/loader.js:600:32)
```

Butt4cak3: What version of node are you running?

server/infoextractor.js
Butt4cak3: That's sort of where it was thrown before. I just changed `video.service === "googledrive"` to `video.mime` to make it independent of the service.

https://github.com/dyc3/opentogethertube/blob/fcb210d6021ac73aea1e97039114e2aba5f31f7d/infoextract.js#L152-L161

I'll see if I can find a better place to throw it.

server/services/dailymotion.js
dyc3: `v10.15.3`, but I also tried with `v12.18.3` and I got the same thing. When did this get added to node?

Butt4cak3: > When did this get added to node?

I can't find anything referencing that it did get added to node. In fact, all I can find is documentation saying that static class properties are not yet implemented. So I tried to run the code in 2 more environments, both are separate Linux boxes (I developed on a Windows PC so far) and I always get the same result. I used node v12.18 in all cases.

I also made a mini test script, which also works wherever I try to run it. I also tried this one in a `node:lts` docker container. Same result.

```js
class TestClass {
  static foo = "bar";
}

console.log(TestClass.foo);
```
```
$ node --version
v12.18.2
$ node test.js
bar
```

The thing is that, even though I don't seem to have a problem using them in node lts, static class properties appear to be not yet officially implemented. The `ServiceAdapter` classes don't really depend on `SERVICE_ID` to be a static member. I made it one for convenience in case we need it further down the line but I didn't use it as such yet. So I guess I'll just make it an instance member to be safe.

server/serviceadapter.js
Butt4cak3: The ServiceAdapter methods now throw `IncompleteServiceAdapterException`s.

Butt4cak3:
I made a good bit of progress today. I changed a few things that you mentioned and I almost fully implemented the YouTube adapter. It now supports searches and linking to videos with and without list parameters, playlists and channels.

With the YouTube adapter mostly out of the way, I can now focus on implementing the other adapters, writing tests and cleaning everything up.

I suggest that we handle pagination in a separate PR and focus on feature parity in this one. This whole thing is pretty much blocking everything related to the server side, because it will probably cause *tons* of merge issues if we modify anything on the backend. That's why I would say that it's better not to make this bigger than it already is.

I managed to get my API key working btw. I don't know what it was, but I resolved it by creating a completely new project in the Google API Console and generating a new key there.

Butt4cak3:
@dyc3 I was testing Google Drive and couldn't get it to work. Then I realized that it doesn't work on the master branch or on the official website either. The Drive API responds with video data, but the response doesn't include `thumbnailLink` or `videoMediaMetadata`. Is that known and would you mind trying it out on your end? I couldn't find a GH issue regarding this.

dyc3:
Yeah, google drive is super finicky. On the official site, I had forgotten to add the api key, whoops. After that, it worked for me.

Butt4cak3:
So this is taking me longer than I expected initially. As you can see, I've mostly written tests during my last couple of programming sessions. The tests aren't complete yet - and admittedly, some of the YouTube tests are pretty bad and should probably rewritten imo - but I'm getting there. Some of the code in the `YouTubeAdapter` also seems pretty messy. But there is one thing that's really bothering me.

My goal was to separate service-specific logic into components and have the info extractor deal with delegating and cache. The `YouTubeAdapter` needs to cache channel playlist IDs in order to optimize its API usage. That's very specific to how the YouTube API works, so it's not something that you could add to the info extractor, so it makes sense to put it in the adapter class. But now the adapter depends on a cache of some sort. The other adapters are really just translation layers for their corresponding APIs and could theoretically be used outside of this project, while the adapter for YouTube is now more than that.

Maybe I'm overthinking this and it really isn't an issue at all because the adapters simply won't get used outside this context. Maybe there's an obvious solution that I'm overlooking. I don't know. I'm just documenting this in hopes that you (or anyone, really) replies with "Hey, why don't you just *x*?".

Either way I'm going to get this PR into a somewhat mergeable state over the next couple of days so prepare for a bit of code review.

antoinebou13:
When this pull request is finished I will try to add an adapter for Spotify

dyc3:
> some of the YouTube tests are pretty bad

Yeah, some of them were written when I was first learning how to use jest.

> The YouTubeAdapter needs to cache channel playlist IDs in order to optimize its API usage.  That's very specific to how the YouTube API works

I don't think it's completely realistic to _completely_ separate caching from adapters. But we can separate everything that all the adapters have in common.

dyc3:
Hey @Butt4cak3, do you plan on finishing this soon? I've intentionally tried to avoid messing with the info extractor until this was done, but we gotta get this merged in.

Butt4cak3:
I planned to have this done AGES ago. I kinda got burnt out from my programming day job and didn't feel like working on side projects at all for a while. That was a while ago but I didn't really get around to this since. I have the day off tomorrow so I'll use that to get back into it and then I'll hopefully be able to finish most of it this weekend. For real this time.

Sorry for disappearing like that.

Butt4cak3:
Two things: I didn't have as much time as I would have liked last weekend and I underestimated the amount of work that's remaining on this PR.

So far I've been trying to get as high of a test coverage as possible on everything. If you're okay with it, then I'd drop that to a state where it's "good enough". Then I can get the room manager done and we can finally merge this PR (that's been open for way too long - again, I'm sorry) to unblock everything else that still needs to be done. We can always add unit more unit tests later. What do you think about that?

dyc3:
Sounds like a plan @Butt4cak3. I can help resolve any merge conflicts and making sure we don't regress, so don't worry about that.

Butt4cak3:
I still can't seem to get Google Drive or direct links to work locally. The player keeps telling me that "No compatible source was found for this media.". However, I can tell that the API calls to single Drive videos work correctly, as I can see thumbnails and titles in the previews and all other necessary information in the CachedVideos table. API requests for folders keep failing for some reason. The code as it is should work™. You're going to have to try it out yourself.

Butt4cak3:
I removed the InfoExtractor unit tests for now because the functionality they tested are mostly covered in the adapter tests. What's left now is the caching and we should probably add tests for that quite soon. Otherwise a cache bug could slip and absolutely destroy API quotas.

Butt4cak3:
Alright. I got this branch up-to-date with master with just a few merge conflicts. One of them was caused by the workaround in 6708b19bcb6ff0cfdef657ee6f01da63c234b67f which I reimplemented in the YouTube service adapter. I tested it with the URLs provided in #285 and made sure the cache works correctly so that the scraping doesn't have to be done more often than necessary.

If I'm not missing anything, this should be at a point where you can test it for regressions. I'll be available all weekend to add and fix stuff to finally get this on its way to get merged.

Once this is somewhat finalized, I'd suggest I/we add some documentation for how to write service adapters to the project wiki.

dyc3:
Fantastic! I'm gonna see if I can spin up a heroku instance for this so I can bring in other people to help test.

dyc3:
Youtube playlists in the add preview appear to not have `length` set, however when the videos are queued up and played it works fine.

Butt4cak3:
From what I can tell, OTT makes *two* requests to the YouTube API for every playlist search.

1. Get the snippet part of every video in the playlist. That includes the ID and thumbnails.
2. Get the complete video info for all video IDs returned by the playlist query.

Is that correct? Because that's what I did in the most recent commit.

dyc3:
Yup, that's correct. It's because youtube doesn't return the video length when you list the playlist items, which is critical for us to have.

server/infoextractor.js
```diff
@@ -0,0 +1,271 @@
...
+
+async function getManyVideoInfo(videos) {
+  const grouped = _.groupBy(videos, "service");
+  const results = await Promise.all(Object.entries(grouped).map(async ([service, serviceVideos]) => {
+    // Handle each service separately
+    const cachedVideos = await storage.getManyVideoInfo(serviceVideos);
+    const requests = cachedVideos
+      .map(video => ({
+        id: video.id,
+        missingInfo: storage.getVideoInfoFields(video.service).filter(p => video[p] == null),
```
dyc3, line 167:
```suggestion
        missingInfo: storage.getVideoInfoFields(video.service).filter(p => video[p] === null),
```

server/infoextractor.js
```diff
@@ -0,0 +1,271 @@
...
+  new GoogleDriveAdapter(process.env.GOOGLE_DRIVE_API_KEY),
+  new VimeoAdapter(),
+  new YouTubeAdapter(process.env.YOUTUBE_API_KEY, redisClient),
+  new DirectVideoAdapter(),
+];
+
+const ADD_PREVIEW_SEARCH_MIN_LENGTH = 3;
+
+function isURL(str) {
+  return URL.parse(str).host != null;
```
dyc3, line 27:
```suggestion
  return URL.parse(str).host !== null;
```

server/infoextractor.js
```diff
@@ -0,0 +1,271 @@
...
+ * @param {string} videoId
+ * @returns {[string, string[]]}
+ */
+async function getCachedVideo(service, videoId) {
+  try {
+    const result = await storage.getVideoInfo(service, videoId);
+    const video = new Video(result);
+    const missingInfo = storage
+      .getVideoInfoFields(video.service)
+      .filter(p => video[p] == null);
```
dyc3, line 44:
```suggestion
      .filter(p => video[p] === null);
```

dyc3 (approved):

server/infoextractor.js
Butt4cak3: The `== null` and `!= null` checks are intended to be double-equals because they cover both `null` and `undefined` cases. It's actually quite common to do this afaik. eslint even has a rule to allow double-equals comparisons with `null`.

PR #306, opened 2020-10-07
Hide scrollbar when in fullscreen

### LINKED ISSUE #273: Small problem with Fullscreen Scrollbar
## Current Behavior
If you go into fullscreen mode (with F or Fullscreen button from the site) the scrollbar is still visible. It's not a hard bug, but it's not real fullscreen either :D

### How To Reproduce
Just go into fullscreen mode. Little css tweaks and it's fixed :D

## Expected behavior
Scrollbar should be disabled or invisible (if you want to allow the user to scroll like in youtube videos) in fullscreen.

## Environment
- [x] This happens on the official site: opentogethertube.com

### Desktop (please complete the following information):
- OS: Windows 10
- Browser: Firefox & Chrome
- Version: FF 77.0.1 | Chrome 83.0.4103.106

elenirotsides:
I was able to remove the scrollbar when the fullscreen event is triggered, but can't seem to add it back after the user exits fullscreen. I tried removing the class, but that didn't work, so then I tried to override the class with another, and that didn't work either. Its very possible that I wasn't doing these things correctly, so I'll take any advice, input, and suggestions anyone can give me! Thanks!

closes #273

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -187,5 +187,8 @@ export default {
 .scrollbarBeGone {
   -ms-overflow-style: none; // I think this is an old way to do this? Probably not ideal
   scrollbar-width: none;
+  &::-webkit-scrollbar {
+    display: none;
+  }
 }
 </style>
--- a/src/App.vue
+++ b/src/App.vue
@@ -180,10 +180,6 @@ export default {
   opacity: 0.7;
 }
 
-.scrollbarBeGone::-webkit-scrollbar {
-  display: none;
-}
-
 .scrollbarBeGone {
   -ms-overflow-style: none; // I think this is an old way to do this? Probably not ideal
   scrollbar-width: none;
--- a/src/App.vue
+++ b/src/App.vue
@@ -128,9 +128,11 @@ export default {
     document.addEventListener('fullscreenchange', () => {
       if (document.fullscreenElement) {
         this.$store.state.fullscreen = true;
+        document.fullscreenElement.classList = "scrollbarBeGone";
       }
       else {
         this.$store.state.fullscreen = false;
+        document.querySelector("html").classList.remove("scrollbarBeGone");
       }
     });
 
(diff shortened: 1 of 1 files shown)


elenirotsides:
Fixed it, should be good to go!

src/App.vue
```diff
@@ -128,9 +128,11 @@ export default {
     document.addEventListener('fullscreenchange', () => {
       if (document.fullscreenElement) {
         this.$store.state.fullscreen = true;
+        document.fullscreenElement.classList = "scrollbarBeGone";
```
dyc3, line 131: It's probably not a good idea to override the classes on the element. It'd be an easy place for new bugs to occur.
```suggestion
        document.querySelector("html").classList.add("scrollbarBeGone");
```

src/App.vue
```diff
@@ -177,4 +179,13 @@ export default {
...
 }
+
+.scrollbarBeGone::-webkit-scrollbar {
+  display: none;
+}
+
+.scrollbarBeGone {
+  -ms-overflow-style: none; // I think this is an old way to do this? Probably not ideal
+  scrollbar-width: none;
+}
```
dyc3, line 190:
```suggestion
.scrollbarBeGone::-webkit-scrollbar {
  display: none;
}

.scrollbarBeGone {
  -ms-overflow-style: none; // I think this is an old way to do this? Probably not ideal
  scrollbar-width: none;
  &::-webkit-scrollbar {
    display: none;
  }
}
```

dyc3 (approved):
It's good, just needs a couple of changes.

src/App.vue
```diff
@@ -177,4 +179,16 @@ export default {
 .text-muted {
   opacity: 0.7;
 }
+
+.scrollbarBeGone::-webkit-scrollbar {
+  display: none;
+}
+
```
dyc3, line 186:
```suggestion
```

dyc3 (approved):

PR #316, opened 2020-11-07
disabled duplicates in the queue

### LINKED ISSUE #20: Duplicate videos can be added to the queue
I can't think of a reasonable use case where a user would want to add duplicate videos to the queue.

elenirotsides:
Disabled the ability to add duplicate videos to the queue. I did this by comparing the video ids in the created hook, and then disabled the button if the video was a duplicate so the addToQueue function couldn't be triggered. Closes #20﻿

### DIFF
--- a/api.js
+++ b/api.js
@@ -43,6 +43,19 @@ function handleGetRoomFailure(res, err) {
 	}
 }
 
+function handlePostVideoFailure(res, err) {
+	/*
+	not sure what to do with this function, I know that I'm supposed to call this function in 
+	the router.post("/room/:name/queue" ..... route but I don't know where to do it
+	*/
+	if (err.name === "VideoAlreadyQueuedException") {
+		res.status(400).json({
+			success: false,
+			error: "Video already in queue",
+		});
+	}
+}
+
 // eslint-disable-next-line no-unused-vars
 module.exports = function(_roommanager, storage) {
 	const roommanager = _roommanager;
--- a/api.js
+++ b/api.js
@@ -44,10 +44,6 @@ function handleGetRoomFailure(res, err) {
 }
 
 function handlePostVideoFailure(res, err) {
-	/*
-	not sure what to do with this function, I know that I'm supposed to call this function in 
-	the router.post("/room/:name/queue" ..... route but I don't know where to do it
-	*/
 	if (err.name === "VideoAlreadyQueuedException") {
 		res.status(400).json({
 			success: false,
@@ -313,14 +309,14 @@ module.exports = function(_roommanager, storage) {
 					res.json({
 						success,
 					});
-				});
+				}).catch(err => handlePostVideoFailure(res, err));
 			}
 			else if (req.body.service && req.body.id) {
 				room.addToQueue({ service: req.body.service, id: req.body.id }, req.session).then(success => {
 					res.json({
 						success,
 					});
-				});
+				}).catch(err => handlePostVideoFailure(res, err));
 			}
 			else {
 				res.status(400).json({
--- a/api.js
+++ b/api.js
@@ -47,7 +47,14 @@ function handlePostVideoFailure(res, err) {
 	if (err.name === "VideoAlreadyQueuedException") {
 		res.status(400).json({
 			success: false,
-			error: "Video already in queue",
+			error: err.message,
+		});
+	}
+	else {
+		log.error("Unhandled exception when getting video:", err);
+		res.status(500).json({
+			success: false,
+			error: "Failed to get video",
 		});
 	}
 }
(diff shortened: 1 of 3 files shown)


src/components/VideoQueueItem.vue
```diff
@@ -50,8 +50,23 @@ export default {
...
 		};
-	},
+	}, 
+	created() { 
+        if (this.item.id === this.$store.state.room.currentSource.id) {
+			this.isLoadingAdd = false;
+			this.hasBeenAdded = true;
+			this.existsInQueue = true;
+		}
+		for (let i = 0; i < this.$store.state.room.queue.length; i++) {
```
dyc3, line 62: Use a `for...of` loop here instead

src/components/VideoQueueItem.vue
```diff
@@ -50,8 +50,23 @@ export default {
...
 			thumbnailHasError: false,
+			existsInQueue: false,
 		};
-	},
+	}, 
+	created() { 
+        if (this.item.id === this.$store.state.room.currentSource.id) {
+			this.isLoadingAdd = false;
+			this.hasBeenAdded = true;
+			this.existsInQueue = true;
```
dyc3, line 60: should return early

src/components/VideoQueueItem.vue
```diff
@@ -50,8 +50,23 @@ export default {
 			isLoadingVote: false,
 			hasBeenAdded: false,
 			thumbnailHasError: false,
+			existsInQueue: false,
 		};
-	},
+	}, 
+	created() { 
+        if (this.item.id === this.$store.state.room.currentSource.id) {
+			this.isLoadingAdd = false;
```
dyc3, line 58: not needed

src/components/VideoQueueItem.vue
```diff
@@ -50,8 +50,23 @@ export default {
...
+			this.hasBeenAdded = true;
+			this.existsInQueue = true;
+		}
+		for (let i = 0; i < this.$store.state.room.queue.length; i++) {
+			if (this.item.id === this.$store.state.room.queue[i].id) {
+				this.isLoadingAdd = false;
```
dyc3, line 64: not needed

dyc3 (changes requested):
`existsInQueue` is functionally equivalent to `hasBeenAdded`, so use that instead. The video's `service` is not being checked either. A video on 2 different services _could_ have the same id. (It's an extremely rare edge case, but one that is easily avoided)

(Also, make sure you don't have trailing or inconsistent whitespace. I fixed it for you this time tho)

src/components/VideoQueueItem.vue
dyc3: The loop should also return early

src/components/VideoQueueItem.vue
elenirotsides: What do you mean should return early?

src/components/VideoQueueItem.vue
elenirotsides: I added a break, lmk if I didn't do this correctly though

elenirotsides:
@dyc3 Can you take a look at my latest commit when you get a chance, if you don't mind? I think I'm on the right track, but I don't know how to finish off my changes in `api.js`. Also, I'm not 100% confident that what I did is in the correct location either.

src/components/VideoQueueItem.vue
dyc3: It should `return` here, because the for loop doesn't need to run

roommanager.js
```diff
@@ -225,6 +225,10 @@ class Room {
 				this.log.error(`Failed to get video info: ${err}`);
 				queueItem.title = queueItem.id;
 			}).then(() => {
+				if (this.queue.id.includes(queueItem.id) && this.queue.service.includes(queueItem.service)) {
```
dyc3, line 228: The condition should look more like: (https://lodash.com/docs/4.17.15#find)
```suggestion
				if (_.find(this.queue, queueItem)) {
```

And this check should go _before_ `getVideoInfo`

api.js
```diff
@@ -43,6 +43,19 @@ function handleGetRoomFailure(res, err) {
 	}
 }
 
+function handlePostVideoFailure(res, err) {
+	/*
+	not sure what to do with this function, I know that I'm supposed to call this function in 
+	the router.post("/room/:name/queue" ..... route but I don't know where to do it
+	*/
```
dyc3, line 50: add `.catch(err => handlePostVideoFailure(res, err));` after the `then` calls

roommanager.js
```diff
@@ -743,6 +746,13 @@ class RoomNameTakenException extends Error {
 	}
 }
 
+class VideoAlreadyQueuedException extends Error {
+	constructor(video) {
+		super(`The video "${video}" is already in the queue`);
```
dyc3, line 751:
```suggestion
	constructor(title) {
		super(`The video "${title}" is already in the queue`);
```

dyc3 (approved):

PR #329, opened 2020-11-18
more test coverage

dyc3:


Butt4cak3:
I'll just use this PR to ask a question if you don't mind. Now that you had to deal with the new InfoExtractor and the service adapters, how do you like the interface? Is it easy enough to work with? I already have a few small things I'd like to change about it, but I'd first like to know if there are any major flaws that need to be addressed first.

dyc3:
It's a 1000x better than what we had before! But I think I'd have to implement a new info extractor to know.

The only problem I have with it is pretty minor: `getCachedVideo` can throw `UnsupportedMimeTypeException`. I don't think this makes a lot of sense, and the mime type check should happen somewhere else.

PR #331, opened 2020-11-20
Handle private playlists in /watch URLs with list parameter

### LINKED ISSUE #330: Youtube links from user's liked videos don't work
## Current Behavior
Links like:
```
https://www.youtube.com/watch?v=MVUWcX1aZnk&list=LL&index=133&ab_channel=BrianJordanAlvarez
```
don't work in the add preview

### How To Reproduce

1. Put the link in the add preview box.

## Expected behavior
Add preview should return 1 video.

## Environment

- [x] This happens on the official site: opentogethertube.com
- [x] This happens using a self-hosted version.

Butt4cak3:
When merged, this pull request adds checks for private playlists to the YouTube service adapter.

When the user searches for a video via a URL that links to a specific video but has a `list` parameter, the adapter will first try to return all videos in that playlist. When any error occurs during that, it will fall back to fetching only that one video's information.

When the `list` parameter in the URL is "WL" or "LL", the adapter will ignore them to save time and API quota.

~~The PR also includes a tiny fix for a potential bug where `resolveURL` didn't always return an array. This usually doesn't come into effect because `resolveURL` only gets called when we're dealing with collections, but this might change in the future.~~ (Reverted)

Fixes #330

### DIFF
--- a/server/services/youtube.js
+++ b/server/services/youtube.js
@@ -14,6 +14,8 @@ const storage = require("../../storage");
 
 const log = getLogger("youtube");
 
+const knownPrivateLists = ["LL", "WL"];
+
 class YouTubeAdapter extends ServiceAdapter {
   constructor(apiKey, redisClient) {
     super();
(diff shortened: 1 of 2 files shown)


Butt4cak3:
Oh well. I broke a test with this. I'll fix it and also add tests for the LL and WL playlists.

Butt4cak3:
Done. Force-pushed to get rid of a reverted commit.

dyc3 (approved):

Issue #253, opened 2020-04-17
Service support:  Spotify


antoinebou13:
Name of service: SPOTIFY
URL of homepage: https://open.spotify.com

Sample video URLs
1. https://open.spotify.com/playlist/1clDgwS8K0EEGXz1iUS0DN?si=kPWg2VU7SS6k94wzrLbKag
2. https://open.spotify.com/album/3tQd5mwBtVyxCoEo4htGAV?highlight=spotify:track:119c93MHjrDLJTApCVGpvx
3. https://open.spotify.com/show/3IM0lmZxpFAY7CwMuv9H4g?si=yQkN2vFLS4uRP7OAHWJdhg
4. https://open.spotify.com/episode/1HoL8UyIxFKvrxLmVBqrpU?si=0Jvi0xifSVGgUJZ71qoIRQ

- [X] Service has a public API
    - [X] Service requires an API key
- [X] Service provides it's own iframe embed
    - [X] Video playback can be controlled via javascript

React with 👍 if you are interested in this.

dyc3:
Ok, so there are 2 options for implementing this, both of which require the user to authenticate with a premium spotify account. 

1. Remote control playback to the user's devices.
2. Streaming the playback in the browser (https://developer.spotify.com/documentation/web-playback-sdk/)

dyc3:
I've already partially implemented link parsing for this in the `spotify` branch.

dyc3:
@antoinebou13 Now that service adapters are merged in, are you still interested in doing this?

antoinebou13:
Sure i can try doing it

antoinebou13:
@dyc3 i love the tests song btw

antoinebou13:
https://github.com/antoinebou13/opentogethertube/pull/1

dyc3:
You should make the PR over here and not on your fork

antoinebou13:
I am not finishing with the PR. I will put it here when I am finishing

dyc3:
We can mark it as a draft until you're done

PR #337, opened 2020-11-25
Service support: Spotify

antoinebou13:
Service support: Spotify #253


Butt4cak3:
I'll take a look at this and try to do a first review pass tomorrow. I don't use Spotify, though, so I can't actually test if it works or not.

dyc3 (commented):
I took a quick look, and this seems like a good start.

- Try to avoid adding packages to `package.json` unless it's absolutely necessary.
- Remember to run `npm run lint`
- Avoid refactoring unless it's absolutely necessary
- There's a few stray changes that aren't relevant to the PR

server/services/spotify.js
```diff
@@ -0,0 +1,114 @@
...
+      title: result.data.name,
+      description: `${result.data.type} ${result.data.name} ${Math.floor((result.data.duration_ms / 1000 / 60) << 0) + ':' + Math.floor((result.data.duration_ms / 1000) % 60)}`,
+      thumbnail: result.data.album.images[2].url,
+      length: result.data.duration_ms,
+    });
+    return video;
+  }
+
+  get isCacheSafe() {
+    return false;
```
Butt4cak3, line 105: Can songs on Spotify just change without getting a new ID?

Butt4cak3 (commented):
I looked at the server code and so far it looks pretty good apart from two little things. Again, I don't use Spotify so I don't think I can test this without a login. But I'm happy to see that the adapter concept seems to work so far.

server/services/spotify.js
dyc3: Metadata can change, but it's not frequent, because its not that easy to do. We can be pretty safe assuming song length never changes though.

antoinebou13: I don't think the metadata change very often, but I am not really sure what to put in this function.

dyc3: I don't think the ID can change because otherwise it would break the spotify camera codes that they generate

package.json
```diff
@@ -27,6 +27,7 @@
     "express": "^4.17.1",
     "express-rate-limit": "^5.1.1",
     "express-session": "^1.17.0",
+    "js-base64": "^2.5.2",
```
antoinebou13, line 30: remove this package and use `new Buffer(this.clientId + ":" + this.clientSecret).toString('base64')`

server/services/spotify.js
Butt4cak3: @antoinebou13 Assuming we can cache the metadata, you can let it `return true;` or, even better, remove it completely.

antoinebou13:
@dyc3 @Butt4cak3 can I call an adapter function inside the player vue component. Like write the API call inside the adapter and use it in the vue

dyc3:
No, the service adapters are server side only.

antoinebou13:
@dyc3 Is there a general router that I can add an endpoint for user authorization for the API. Currently only a router for the room and API, but it's not really generic to add a new API endpoint easily
https://developer.spotify.com/documentation/general/guides/authorization-guide/#implicit-grant-flow

dyc3:
Right now, on the server side, `api.js` _is_ the general router. Service adapters can't add api endpoints, and I don't think that would be a good idea to make possible.

What do you need to do that requires an api endpoint? Do we need to handle spotify logins? Do we need to implement spotify oauth?

antoinebou13:
The user has to accept that opentogethertube modify the user music playback and the API requires to have a callback url. I just need a callback endpoint to receive the access_token to save the information inside the cookies or local storage to access the Spotify player API to modify music playing for the current user in opentogethertube .

dyc3:
You should be able to create a new callback in `api.js`. In the callback, you can feed the access token to the room manager, who can distribute it to the client via websocket.

api.js
```diff
@@ -469,5 +469,10 @@ module.exports = function(_roommanager, storage) {
 		});
 	});
 
+	/* Spotify Authentication Callback url */
+	router.get("/spotify/authentication", (req, res) => {
+		res.redirect(`${req.query.state}?${req.query.code}`);
```
antoinebou13, line 474: @dyc3 not really sure how to send the data with the websocket

dyc3: You can use the session from the request (`req.session`) to compare it with clients connected to rooms (`roommanager.js`)

dyc3:
Hey @antoinebou13, how's this PR coming along? Are you running into any problems?

antoinebou13:
I am almost done with a working prototype, but I have an issue with play and seek in the player (I will send you a video of the issue if I can not find a solution). I still have to the server auth with the feedback you gave me. I still have to do the playlist parsing for the queue. Also, I would like to add a flag to disable Spotify in the config.

dyc3:
Sounds great!

> I would like to add a flag to disable Spotify in the config.

Shouldn't be a problem. Remember to update the documentation in `docs/config.md`.

Butt4cak3:
I don't know about the Spotify config, but I found something that seems weird to me. Is the client *secret* supposed to show up in client code?

antoinebou13:
You rigth i will to hide and put it the server then the client request it as base64

Butt4cak3:
If a user isn't supposed to see the secret then that won't help. I don't know how to actually use the Spotify API. I just noticed this and wondered.

dyc3:
I just realized something. You know how we use passport for discord login? What if we do the same thing for spotify, and allow users to create/link OTT accounts with spotify. I realize that doing this would change a lot about how this PR works, but what do you guys think?

antoinebou13:
What I like with your application/web is the plug-and-go features and the main reason I want the Spotify service adapter is that I can use OTT if I want to listen to music for multiple services with my friends without login into a website. I think you are right it would be better to link a Spotify account to an OTT user. Spotify API is flexible. So we can link a user session or cookie to a token then use it as we want. I currently working to hide the client secret and learning more about express.

maddie-wang:
Thoughts on using this approach for spotify?

- paste in the playlist link

- we throw the metadata into YouTube, and play those songs


https://groovy.zendesk.com/hc/en-us/articles/360030600932-How-Groovy-plays-music-from-Spotify-and-Apple-Music-Deezer-

dyc3:
That's an interesting approach, @maddie-wang . I think it's worth investigating.

maddie-wang:
Yay! I thought it might be nicer because it doesn't require people with a Spotify Premium account to use the music sync. The music would be accessible for everyone, not cost-prohibitive.

It's also beneficial since I imagine many users already formed awesome playlists on Spotify. I imagine it can be more convenient to use those playlists than to manually create one every time on Open Together!

dyc3:
@maddie-wang would you be interested in submitting a separate PR for it?

maddie-wang:
Sure!

antoinebou13:
@maddie-wang You can use what a already have and change it. I had issues with the sync of the time between clients, but with your method, you skip that entirely and it's would be much simpler for the user and the implementation of the logic and also skip premium.
pseudo code
```
const result = await axios({
      url: https://api.spotify.com/v1/playlists/{playlist_id}/tracks,
      method: 'post',
      data: "grant_type=client_credentials",
      headers: {
        'Authorization': `Basic ${Buffer.from(this.clientId + ":" + this.clientSecret).toString('base64')}`,
      },
    }).catch(function (error) {}
// local
song_names = []
for(item in result.items){
song_names.push(item.track.name)
}
```

dyc3:
I think this branch is a little too far behind, it probably would be difficult to bring it up to date. Closing for now, feel free to make a new PR.

PR #345, opened 2020-12-22
Custom Video Slider

### LINKED ISSUE #319: Refactoring: New Slider Component
The current slider from the `vue-slider` package is a relic from when I was first prototyping the website. It's worked great so far, but it has started to get in the way of UX. We need to create a new slider that meets the following requirements:

- Must not use `this.$store` at all
- Must be able to display value when the mouse is hovered over different parts of the bar in a tooltip
    - Like a youtube video's slider
    - Should not be default behavior, opt in with a boolean prop
- Must be able to customize string format of slider value tooltip with a function.
- Must emit `dragStart`, `dragChange`, `dragEnd` events when the user drags the slider
- Must be able to highlight different sections of the slider with different colors, independent of the slider value.
   - For example, the 0-5 range could be marked blue, and the 13-20 range could be marked green.

elenirotsides:
Opening up this PR as a draft to start, so I can break up my progress into tiny fragments while also enabling collaboration and feedback. Right now, I have a 'VideoSlider' component that outputs a 'Hello World' string in Room.vue. Closes #319

### DIFF
--- a/api.js
+++ b/api.js
@@ -27,6 +27,11 @@ const VALID_ROOM_QUEUE_MODE = [
 	"vote",
 ];
 
+const VALID_ROOM_TOOLTIP = [
+	"none",
+	"timestamp",
+];
+
 function handleGetRoomFailure(res, err) {
 	if (err.name === "RoomNotFoundException") {
 		res.status(404).json({
@@ -102,6 +107,7 @@ module.exports = function(_roommanager, storage) {
 				"isTemporary",
 				"visibility",
 				"queueMode",
+				"tooltip",
 				"queue",
 				"clients",
 			]));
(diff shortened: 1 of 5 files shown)


dyc3:
I noticed you committed `package-lock.json`. Could you remove that from your commit?

elenirotsides:
Ahh yes, my bad. I think it updated some of the dependancies for some reason unbeknownst to me; I didn't actually touch that file but I'll make sure to be more careful when I push stuff up now!

elenirotsides:
Whoops...I hit the wrong button before on my keyboard....never meant to close this!

dyc3 (commented):
I noticed you added a room setting for a tooltip. This doesn't make much sense, because why should everyone in the room have the ability to enable/disable the tooltip? And why the tooltip in particular?

This new slider should be a completely isolated component, that does not rely on external variables. When the slider is used, parameters will be provided via `props` on the component. **There shouldn't be any changes to server side code.** I guess I should have been more specific.

https://vuejs.org/v2/guide/components-props.html

dyc3:
Closing for inactivity.

PR #470, opened 2021-08-13
New logo

Leomm04:
I created a new logo that fits a bit more into the design


dyc3:
Thanks for the PR! I wasn't really looking for a logo change, but I'm open to discuss it.

Github doesn't really show previews for images. Can you edit the PR to include an image preview of the new logo? Also, I'd like to have an svg of the logo rather than multiple copies of the same logo in multiple resolutions. If we really need that, I would prefer to have those be generated automatically with a build script.

Leomm04:
> 
> 
> Thanks for the PR! I wasn't really looking for a logo change, but I'm open to discuss it.
> 
> Github doesn't really show previews for images. Can you edit the PR to include an image preview of the new logo? Also, I'd like to have an svg of the logo rather than multiple copies of the same logo in multiple resolutions. If we really need that, I would prefer to have those be generated automatically with a build script.

Here's the image: https://raw.githubusercontent.com/dyc3/opentogethertube/0d4feccef7ad94d922bfb2d60cba9fd347ca2442/src/assets/logo2048.png.

I can send you the PDN file for it.

dyc3:
Cool design! However, I'm not sure if I want to move forward with this design. I don't really see how this fits with the current design better, and I would prefer to not make such a drastic change to the logo's overall shape.

Leomm04:
> 
> 
> Cool design! However, I'm not sure if I want to move forward with this design. I don't really see how this fits with the current design better, and I would prefer to not make such a drastic change to the logo's overall shape.
> 
> It's a lot better to discuss these kinds of changes before making a PR so you don't do a bunch of work for something that won't be used.

No problem. I just made it, because I was bored. I'll probably use it as a logo in my project.

Leomm04:
No problem. I just made it, because I was bored. I'll probably use it as a logo in a project if I make one, that fits it.

PR #484, opened 2021-09-16
small log fix

dyc3:
- update typescript, improve null checking
- add some more null checks, rework most events typing to make it easier
- add some more null checks, rework undo request typing a little
- small logging fix


dyc3:
accidentally included some extra commits

PR #530, opened 2021-10-13
Localization infrastructure

### LINKED ISSUE #205: Localization
According to my analytics, we should prioritize

- German
- Dutch
- French
- Italian

Nice to have

- Spanish
- Russian
- Korean
- Chinese
- Japanese

I have do idea how this should be implemented. I'm open to comments and pull requests.

dyc3:
- add i18n plugin for localization closes #205
- replace some strings with localizable versions

Introduces bare minimum infrastructure for translating the UI into different languages.

Sufficient to close #205 in favor of opening up more granular issues.


dyc3:
Unfortunately, I think this is blocked until we completely separate out the client and server into their own modules.

dyc3:
This one probably needs to be redone now.

PR #631, opened 2022-02-14
Chat rework

### LINKED ISSUE #622: Chat in Fullscreen Mode

OpenTogetherTube is almost the perfect replacement for Metastream.
Unfortunately there's no way to use or see the chat in full-screen mode for the best experience.

I would like to use the chat function in full-screen video mode like it's happening in Metastream with an overlay box you make appear by pressing Enter on the keyboard. Would it be possible to implement that feature?

### Additional context

_No response_

dyc3:
- refactor(Chat): use composition api instead
- move Chat inside video player
- adds hotkey <kbd>t</kbd> to open char, and close it with <kbd>Esc</kbd> while text box is focused.

Closes #622

### DIFF
--- a/client/src/components/Chat.vue
+++ b/client/src/components/Chat.vue
@@ -91,6 +91,8 @@ export default Chat;
 </script>
 
 <style lang="scss" scoped>
+@import "../variables.scss";
+
 .chat-header {
 	border-bottom: 1px solid #666;
 }
@@ -105,7 +107,9 @@ export default Chat;
 
 .message {
 	margin: 4px;
+	margin-left: 0;
 	padding: 3px;
+	padding-left: 0;
 
 	&:first-child {
 		margin-top: auto;
@@ -120,6 +124,11 @@ export default Chat;
 
 	.from {
 		font-weight: bold;
+		margin-left: 0;
+	}
+
+	@media screen and (max-width: $md-max) {
+		font-size: 0.8em;
 	}
 }
 
--- a/client/src/components/Chat.vue
+++ b/client/src/components/Chat.vue
@@ -33,6 +33,7 @@
 				@keydown="onInputKeyDown"
 				v-model="inputValue"
 				autocomplete="off"
+				ref="chatInput"
 			/>
 		</div>
 	</div>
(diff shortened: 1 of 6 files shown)


dyc3:
I've canaried some changes to the chat UI to https://opentogether-chat-rewor-7vf9tf.herokuapp.com/ Looking for feedback. Let me know what you think

Shu2ik:
It would work out as basic idea.
The time the comments stay on screen is good 👍

These things could be optimized
• It's rather small the way it looks, especially in width. Makes it difficult to overview if needed. I think you can double the size and make the window slightly transparent maybe
• More padding could be added to the notification boxes especially on left side
• Sometimes comments get cut off if you post up to 5 or keep posting
• Activation and navigation without mouse is still missing I assume

This is the first impression so far.

Shu2ik:
So putting the text field into the player bar is not within the bound of possibility?

dyc3:
Yeah, looks are not final. Will definitely make the chat and font size bigger, but I need to handle smaller screens more gracefully first. You can activate the chat box with <kbd>t</kbd>, and close it with <kbd>Esc</kbd>. (Should've mentioned that)

> So putting the text field into the player bar is not within the bound of possibility?

I mean it's technically possible, but I'm not going to, mostly because it would dramatically increase complexity.

Shu2ik:
Alright. Guessed so that it's a mock-up more or less.
Overall it's fine, just needs a balanced size ratio between different screens indeed.

Btw, in fullscreen Esc is closing the mode instead of closing the chat.
I expected having the chat box key on 'c' and didn't try 't' which is supposed to stand for text I guess.

dyc3:
Unfortunately, I found out that its not possible to override escape exiting fullscreen. As an alternative, <kbd>Enter</kbd> with no message in the chat box will close the chat without sending a message. I've made all of the other changes though.

I think the last thing I want to change is the location of the user list, but I think I will leave that for another PR.

dyc3:
If there's no remaining feedback, I'm going to merge this in by the end of the day.

Shu2ik:
Looking good and working smoothly. That will do 👍
Opening chat with Enter isn't possible? For consistency.

dyc3:
I think opening chat with Enter has the potential to cause ambiguity and confusion from blind users. Enter usually means "Confirm" or "Send" or "Do this" or "Click". So from an accessibility standpoint, no I don't think its possible.

Shu2ik:
I see. It's a minor detail, no worries.
Thank you for the rework.

Shu2ik:
Will the changes only appear on that specific page or are you going to apply it to the main page as well?
https://opentogether-chat-rewor-7vf9tf.herokuapp.com/

dyc3:
It should have deployed by now, i'll do a manual deploy right now

PR #632, opened 2022-02-14
Fixed layout

xRealNeon:
- Same footer everywhere image
- Removed space under footer 
image


dyc3:
I don't think this is necessary.

Issue #616, opened 2022-02-04
German Localization


xRealNeon:
### Is your feature request related to a problem? Please describe.

Localization

### Describe the solution you'd like

- Adding all strings in the ui to the lang file
- How can a user change the language?
- I can add the missing strings and make a German translation

dyc3:
Thanks for your interest, that would be _super_ awesome! I should probably write some docs for how to add new languages.

There's no UI for switching languages yet (on my todo list). You should be able to add a file called `de.ts` to `client/src/locales/` and as long as you match the schema of `en.ts` it should work. As a temporary thing while developing, you should be able to use the environment variable `VUE_APP_I18N_LOCALE` to set the language.

If you're going to work on this, please separate this in to 2 different PRs:
1. Adding all the remaining strings to `en.ts`
2. Adding German.

xRealNeon:
How do I translate text with links or something like this. @dyc3 

![grafik](https://user-images.githubusercontent.com/20997346/152616543-945be980-98a7-4832-aa34-066bb037fb23.png)

xRealNeon:
I think I should do it like described here: https://vue-i18n.intlify.dev/guide/advanced/component.html

xRealNeon:
but with texts with a lot of new lines this would also be impractical.

dyc3:
Tbh, I'm not completely sure about the best approach. But that article you linked seems pretty good

xRealNeon:
Where are the usernames generated?

dyc3:
> Where are the usernames generated?

They are generated using the `unique-names-generator` package, ctrl + f around for that.

> Is it a good idea to use localized usernames?

No, probably not.

dyc3:
Hey @xRealNeon, hows this coming along?

dyc3:
Unassigned for inactivity

dyc3:
Closing as completed

PR #660, opened 2022-03-23
Add german translation

FriedrichGaming:
This pull request currently adds basic german translation. I'll probably add more translations in the next few days. If you want to merge this pull request already, i can open a new one for further translations.

partially fix for: #616

Also fixes this:
grafik


dyc3 (changes requested):
Please exclude untranslated strings. It will automatically fall back to english.

Otherwise, I can't read German. Rubber stamp LGTM!

FriedrichGaming:
Should I also remove references to other string that weren't changed?
Like for example:
grafik

dyc3:
Yes, that would make sense.

FriedrichGaming:
That should be it. It's currently 5:40am in germany so i'm gonna sleep now.

There are still some inconsistencies regarding the use of formal or informal language but i'll fix those in another pull request.

dyc3 (approved):
Run `yarn lint` to fix the linting errors

PR #669, opened 2022-05-13
make the screen API optional

### LINKED ISSUE #668: Blank page on mobile safari
### Current Behavior

On the page for a room, instead of UI there is just a blank page with the OTT menu bar.

### Expected Behavior

The player and video shows up

### Steps To Reproduce

1. Open OTT
2. Join any room

### Environment

- [X] I'm using the docker image.

### Anything else?

Tested on an iPhone 11

dyc3:
partial fix for #668

### DIFF
--- a/client/src/views/Room.vue
+++ b/client/src/views/Room.vue
@@ -350,7 +350,7 @@ export default {
 
 			i_timestampUpdater: null,
 
-			orientation: screen.orientation.type,
+			orientation: screen ? screen.orientation.type : undefined,
 			videoControlsHideTimeout: null,
 
 			api,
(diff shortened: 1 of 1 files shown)


dyc3:
Unfortunately, heroku seems to be having some issues that's preventing me from creating a review app. @gianklug

dyc3:
I'm going to merge this in because it doesn't seem to have any problems, but there's some other issues that need to be fixed in order to close #668

gianklug:
i just took a look at the code and I think you are following the wrong approach. you are checking if screen exists and then still are trying to access the screen rotation. screen, however always exists, it's just the screen.orientation api that doesn't on iOS.
therefore i still get the same errors ^^'

PR #717, opened 2022-09-14
Add German translation for client-settings

IceBeach7:
Adds German translation for client-settings, and changes the English version to be consistent.


dyc3 (approved):
Thanks for helping keep these strings updated!

dyc3:
If you fix the lints with `yarn lint` you should be good to go.

PR #751, opened 2022-10-14
fixed/improved ru locale

SabatonShip:
Hi! Corrected typos and improved translation quality.


SabatonShip:
изображение
Please pay attention to the display of Russian localization on your mobile device. It was very difficult to find the right translation for the title, so the first word turned out to be very long and, most likely, that's why the text goes off the screen and creates a white line. Maybe it's worth reducing the size of the title on the phone, or should I choose a different translation of the title?

SabatonShip:
> Hmm, yeah I see what you mean. It might be worth making the font size for this configurable by the locale. This is fine for now though, I'll test it later.

Hi! I just checked it on PC and this pull should fix it

PR #772, opened 2022-11-06
Upgrade to Vue3 + Vuetify 3 + Vite, with lots of other major improvements

dyc3:
### TODO

- [x] update docs for how to create a locale
- [x] fix unit tests
- [x] fix CreateRoomForm not validating, making creating permanent rooms impossible
- [x] UserList: fix dropdowns on each user not being on the same line
- [x] switch util/connection to a plugin, and use provide/inject
- [x] switch `util/api` usages to `util/roomapi` composable
- [x] fix timestamp text not updating to the actual timestamp of the video
    - https://github.com/vuejs/core/discussions/7120
- [x] fix the warnings related to the video queue and vuedraggable
- [x] fix "create room" button on blank room list view not working
- [x] fix success style toast showing incorrect icon
- [x] make sure SHORT_URL environment var is respected in prod build
    - had to be renamed to `OTT_SHORT_URL_HOSTNAME`
- [x] fix ShareInvite not having the clipboard icon button anymore
- [x] LoginForm: fix layout
- [x] PermissionsEditor is gone?
- [ ] during testing, sometimes isPlaying state got out of sync and prevented playback from being started on all clients. very weird
    - this might be only fixable by having the server always send an isPlaying sync if it gets a PlaybackRequest
- [x] nav bar buttons are centered for some reason
- [x] make sure captions still work
- [x] nav bar isn't hidden in fullscreen
    - inverted scroll is currently missing in vuetify 3
    - https://github.com/vuetifyjs/vuetify/pull/16108
    - https://github.com/vuetifyjs/vuetify/pull/16267
    - defering fully correct behavior until later: https://github.com/dyc3/opentogethertube/issues/799
- [x] `v-overlay` background color should be black, not white
    - I think the bg colors for the dark theme and light theme need to be swapped 
- [x] debug stuff is visible in prod
- [x] client does not respect position update when the video ends in dj mode until playback state is modified
- [x] RoomSettingsForm: save button does not stick to bottom of screen anymore
- [x] check if robots.txt is present in prod build
- [x] Toast notifications stretch to the size of the longest notification when they shouldn't
- [x] prune unused packages
- [x] Toast notifications with success or error don't have the little close button visible
- [ ] update client unit test eslint config to have vitest rules instead of jest rules
    - might not be worth the effort, [eslint-plugin-vitest](https://github.com/veritem/eslint-plugin-vitest) seems buggy right now
- [x] user registration broke on the server side?
- [x] make sure permanent room creation and saving still works
- [x] `Chat` component tests
    - [x] needs some way to mock the websocket connection
- [x] maybe make sure `OttRoomConnectionMock` doesn't make it into the production bundle?
- [x] Room: convert to composition api (finally)
    - [x] move room control keybinds out of Room view and into a composable?
        - [x] need to reimplement chat activate when you press t
        - [x] it would be really nice to find a way to auto-magically unbind keys when the component is unmounted instead of having to remember to do it myself
        - [x] write unit tests for it
    - [x] getting disconnected and reconnecting while any video is playing causes player to flicker on reconnect
    - [x] loading room settings seems unnecessarily slow
    - [x] if you create a room while already connected to a room it doesn't navigate you to the new room and it exposes an internal error to the user that will be confusing
- [x] fix dockerfiles
- [x] playback buttons and text in the video controls is impossible to see on the light theme
- [x] document how to run cypress e2e and component tests and what differences are
- [ ] it's possible to get into a state where you connect to a room multiple times
    - happens especially when hot reloads occur
    - I think this is a server bug, the room is not getting notified that the clients are disconnecting
- [x] faq page is broken
- [x] Chat: does not force word wrap if first line is longer than chat width
- [x] Chat: does not force sticky to bottom sometimes
- [x] update to cypress 12
- [x] click and drag to re order queue is not working correctly

Follow up items
- [ ] create a global event bus so we can decrease the amount of hacks in the Room view. for example, see comment in `setAddPreviewText`
- [ ] add `plugin:vue/vue3-strongly-recommended` to eslint config
- [ ] refactor server so that jest can be used without babel
    - [ ] maybe requires converting server completely to es module imports in order to use jest without babel?

Pre-merge checklist

- [x] Make sure there are no warnings / errors in the console.
    - It's about as good as its gonna get right now
- [x] Make sure to set `OTT_SHORT_URL_HOSTNAME` in prod
- [x] make sure built site is minified


dyc3:
I feel confident enough to merge this in, but there are going to be some caveats:

- Cypress, when running in the context of CI, is really unreliable, and there is seemingly nothing I can do about it. It really reduces the value of these tests, but at least they're reliable locally.
- The app bar behavior didn't make it into vuetify 3.1: https://github.com/dyc3/opentogethertube/issues/799

PR #829, opened 2023-03-12
Show a button when playback is blocked

### LINKED ISSUE #827: Direct MP4 sources do not launch automatically
### Current Behavior

The mp4 media does not launch automatically on other devices whereas the player of the site is on play. This therefore gives us jerks (due to the syncroinization which works perfectly 👌) but to be able to watch the media we have to right click on the player and click on "Play" so that the player no longer bugs.

### Expected Behavior

This is therefore not a bug but rather an improvement to be made 👍 The reading of the MP4 file should therefore be automatic if the site player is on play.

### Steps To Reproduce

-> Launch an MP4 file on the administrator device
-> Go to the room on another device

### Environment

### Anything else?

This improvement could allow better compatibility especially with smartphones/tablets where we cannot right-click to start playback manually and where the site is unusable on this type of file.

dyc3:
- make it possible to catch when starting playback fails
- show a prompt when media playback is blocked
- fix button styling
- seek to position when unblocking playback
- add waitForPlayer so playback blocked panel appears on page load

fixes #827

### DIFF
--- a/client/src/components/players/DirectPlayer.vue
+++ b/client/src/components/players/DirectPlayer.vue
@@ -55,9 +55,6 @@ export default {
 				type: this.videoMime,
 			});
 			this.player.load();
-			if (this.$store.state.room.isPlaying) {
-				this.player.play();
-			}
 		},
 		beginNewVideo() {
 			this.player = videojs(document.getElementById("directplayer"), {
--- a/client/src/components/players/GenericHlsPlayer.vue
+++ b/client/src/components/players/GenericHlsPlayer.vue
@@ -59,9 +59,6 @@ export default {
 				type: "application/x-mpegurl",
 			});
 			this.player.load();
-			if (this.$store.state.room.isPlaying) {
-				this.player.play();
-			}
 		},
 		beginNewVideo() {
 			this.player = videojs(document.getElementById("generichlsplayer"), {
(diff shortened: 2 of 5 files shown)


Scraft02:
Just to tell you, I am a French user (by the way, sorry for my English mistakes) your site is not known in France and it's a shame because it's so good. I was on Wath2Gether before but I would admit that OpenToGetherTube is much easier to use, beautiful, intuitive and perfectly does the job it is asked to do, which is to synchronize a video with several users) So, I sending you full of strength from my home and thank you for this incredible speed 🤩 I hope the project will continue for a long time 💪

dyc3:
Thank you, @Scraft02! That's very kind.

If you're interested, I would be open to a french localization. The current documentation for how to do that is here: https://github.com/dyc3/opentogethertube/blob/master/docs/how-to-add-new-language.md

dyc3:
Great! I want you to get credit for your work, so can you create a pull request?

Scraft02:
tell me, did you manage to add my translation to the site. If I have to correct something do not hesitate to contact me to tell me what to do. I don't know how to code but I can be the site's French translator 😁 (for future updates and adding text to translate).

dyc3:
Let's keep this discussion on #835

PR #835, opened 2023-03-14
Add French translation

Scraft02:
Translation carried out by me for French users residing in France

### DIFF
--- a/client/src/locales/fr.ts
+++ b/client/src/locales/fr.ts
@@ -94,7 +94,7 @@ export default {
 		"home": "Accueil",
 		"browse": "Parcourir",
 		"faq": "FAQ",
-		"bug": "Rapporter un BUG",
+		"bug": "Rapporter un bug",
 		"support": "Me Soutenir!",
 		"login": "Se connecter",
 		"link-discord": "Lien Discord",
--- a/client/src/locales/fr.ts
+++ b/client/src/locales/fr.ts
@@ -171,8 +171,8 @@ export default {
 			'Recherche Youtube pour "{search}" en appuyant sur entrée, ou en cliquant sur le bouton recherche.',
 		"platforms": {
 			"youtube-videos": "Des vidéos Youtube: {url}",
-			"vimeo-videos": "Des vidéos Viméo: {url}",
-			"dailymotion-videos": "Des vidéos ailymotion: {url}",
+			"vimeo-videos": "Des vidéos Vimeo: {url}",
+			"dailymotion-videos": "Des vidéos Dailymotion: {url}",
 			"any-mp4-videos": "Toutes les vidéos .mp4 publiques: {url}",
 			"youtube-playlists": "Les playlists Youtube: {url}",
 			"youtube-channels": "Les chaînes Youtube: {url}",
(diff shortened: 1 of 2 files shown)


client/src/locales/fr.ts
```diff
@@ -0,0 +1,318 @@
...
+		text: "Création d'une salle temporaire pour vous...",
+	},
+	"attribution": {
+		"sponsorblock-text": "Utilise les données SponsorBlock de",
+	},
+	"nav": {
+		"home": "Accueil",
+		"browse": "Parcourir",
+		"faq": "FAQ",
+		"bug": "Rapporter un BUG",
```
dyc3, line 97: This doesn't seem right, should "BUG" be in all caps here?

Also, google translate has a contributor reviewed translation for "Report a bug" -> "Signaler un bug". What's the difference between "Rapporter un bug" and "Signaler un bug"?

client/src/locales/fr.ts
```diff
@@ -0,0 +1,329 @@
...
+		"single-videos": "Vidéos uniques",
+		"playlists": "Playlists",
+		"playlist": "Playlist",
+		"text": "Ou tapes juste du texte pour faire une recherche Youtube.",
+		"search-for":
+			'Recherche Youtube pour "{search}" en appuyant sur entrée, ou en cliquant sur le bouton recherche.',
+		"platforms": {
+			"youtube-videos": "Des vidéos Youtube: {url}",
+			"vimeo-videos": "Des vidéos Viméo: {url}",
+			"dailymotion-videos": "Des vidéos ailymotion: {url}",
```
dyc3, line 175: typo

client/src/locales/fr.ts
```diff
@@ -0,0 +1,329 @@
...
+		},
+		features: {
+			"title": "Fonctionnalités Principales",
+			"syncronized-playback": {
+				title: "Vidéo synchronisée",
+				text: "Appuies sur le bouton play et la vidéo se lance pour tout le monde dans la salle. Aussi simple que ça.",
+			},
+			"permanent-rooms": {
+				title: "Salles Permanentes",
+				text: "Toi et t'es amis êtes des utilisateurs réguliers ? Tu peux créer une salle permanente qui te permettra de conserver la même URL pour ta salle",
```
dyc3, line 44: Is it normal to have a space before the "?" in french? Seems inconsistent with other punctuation.

dyc3 (changes requested):
Seems good, but I have a few questions.

client/src/locales/fr.ts
Scraft02: Effectively we add a space before a ?  😁 French uses a lot of characters with accents if that's a problem I can remove them but it can make the text less easy to read

client/src/locales/fr.ts
Scraft02: I correct 👍

client/src/locales/fr.ts
Scraft02: Yes indeed for the bug, it is better to put it in lowercase. No difference between "Rapporter un bug" and "Signaler un bug".

client/src/locales/fr.ts
dyc3: Ah, I see. We'll leave it as is then.

dyc3 (approved):

PR #854, opened 2023-04-21
refactor: use redis a bit less

dyc3:
- make it so redis pubsub is not used for room messages
- remove publish, subscribe, unsubscribe calls, scaffold unload event, refactor
- emit unload events
- refactor to use unload event

also includes some more general cleanups


dyc3:
e2e tests pass locally, so merging

PR #953, opened 2023-06-29
Very experimental pluto tv support

dyc3:
- init pluto adapter
- pluto extractor is functional for movies, except playback because CORS
- add a couple test fixtures

~~Currently doesn't work because pluto serves CORS headers that blocks the request in browsers.~~ Fixed by rolling my own CORS proxy.


dyc3:
Turns out that Pluto embeds their ads directly into the hls streams. Interesting.

PR #958, opened 2023-06-30
crank up the retries for cypress to make it more reliable

dyc3:


dyc3:
Nevermind, doesn't work

PR #1070, opened 2023-09-02
WIP: balancer: stream proxy requests instead of buffering them in memory

dyc3:
TODO
- [ ] deal with connection pooling
- [ ] get the outbound response to be streamed instead of just the inbound request


dyc3:
Closed because this is too out of date

PR #1071, opened 2023-09-14
possible fix for flaky e2e component tests

dyc3:


dyc3:
Doesn't seem to fix the problem.

PR #1092, opened 2023-10-04
Disconnect clients when monolith connection is severed

### LINKED ISSUE #1076: Balancer should disconnect clients when the upstream connection to the monolith is severed.
### Current Behavior

The balancer does not do this, and it results in join messages not being resent, and this log message is emitted

```
2023-09-23T16:03:05.045 app[591854ead93383] ewr [info] 2023-09-23 16:03:05 clientmanager error Balancer sent message for client that does not exist or is not a balancer client 
```

### Expected Behavior

The balancer should disconnect all clients when the upstream monolith connection is interrupted to allow them to reconnect. 

Alternatively, it would be possible to avoid the reconnect overhead if the balancer could send joins when the connection is reestablished. But that sounds like a premature optimization.

### Steps To Reproduce

1. Start a monolith
2. Start a balancer
3. Create a room and join it
4. Restart the monolith
5. Send a message in chat
6. See error log message in monolith logs
7. See that chat message does not appear in chat.

### Environment

### Anything else?

The workaround is to have the clients refresh the page.

cjrkoa:
fixes #1076

### DIFF
--- a/crates/ott-balancer-bin/src/balancer.rs
+++ b/crates/ott-balancer-bin/src/balancer.rs
@@ -262,8 +262,23 @@ impl BalancerContext {
         self.monoliths.insert(monolith.id(), monolith);
     }
 
-    pub fn remove_monolith(&mut self, monolith_id: MonolithId) {
+    pub async fn remove_monolith(&mut self, monolith_id: MonolithId) -> anyhow::Result<()> {
         self.monoliths.remove(&monolith_id);
+        
+        self.rooms_to_monoliths
+            .retain(|_, v| *v != monolith_id);
+
+        self.clients
+            .retain(|_, v| self.rooms_to_monoliths.contains_key(&v.room));
+        
+        for client in self.clients.values(){ 
+            client.send(SocketMessage::Message(Message::Close(Some(CloseFrame {
+                code: CloseCode::Library(1006),
+                reason: "Monolith disconnect".into(),
+            })))).await?;
+        }
+        
+        Ok(())
     }
 
     pub fn add_room(&mut self, room: Room, monolith_id: MonolithId) -> anyhow::Result<()> {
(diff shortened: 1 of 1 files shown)


dyc3:
You gotta work on your PR titles.

cjrkoa:
fixes #1076

crates/ott-balancer-bin/src/balancer.rs
```diff
@@ -262,8 +262,23 @@ impl BalancerContext {
...
+            .retain(|_, v| *v != monolith_id);
+
+        self.clients
+            .retain(|_, v| self.rooms_to_monoliths.contains_key(&v.room));
+        
+        for client in self.clients.values(){ 
+            client.send(SocketMessage::Message(Message::Close(Some(CloseFrame {
+                code: CloseCode::Library(1006),
+                reason: "Monolith disconnect".into(),
+            })))).await?;
```
dyc3, line 278: if this fails, it would prevent the disconnects of any subsequent clients. we can't let errors break this for loop.

crates/ott-balancer-bin/src/balancer.rs
```diff
@@ -262,8 +262,23 @@ impl BalancerContext {
...
+        
+        self.rooms_to_monoliths
+            .retain(|_, v| *v != monolith_id);
+
+        self.clients
+            .retain(|_, v| self.rooms_to_monoliths.contains_key(&v.room));
+        
+        for client in self.clients.values(){ 
+            client.send(SocketMessage::Message(Message::Close(Some(CloseFrame {
+                code: CloseCode::Library(1006),
```
dyc3, line 276:
```suggestion
                code: CloseCode::Away,
```

dyc3 (approved):

PR #1111, opened 2023-10-21
add config options to disable room creation

dyc3:


dyc3:
Force merging as test failures are unrelated.

PR #1112, opened 2023-10-21
unicast room messages

### LINKED ISSUE #1100: balancer: Implement unicast room message handling

Currently, all `MsgM2B::RoomMsg`s are handled by simply broadcasting the message to all clients.

Resolve this TODO: https://github.com/dyc3/opentogethertube/blob/98653381d580a99c6890e70e504bdc6bea37a3db/crates/ott-balancer-bin/src/balancer.rs#L558

When `client_id` is `Some`, the message should only be sent to one client with the matching client id.

### Additional context

_No response_

cjrkoa:
fixes #1100
in progress


crates/ott-balancer-bin/src/balancer.rs
```diff
@@ -622,17 +622,23 @@ pub async fn dispatch_monolith_message(
                     };
 
                     // TODO: also handle the case where the client_id is Some
```
dyc3, line 624: nit: remove todo

crates/ott-balancer-bin/src/balancer.rs
```diff
@@ -622,17 +622,23 @@ pub async fn dispatch_monolith_message(
...
-                    // TODO: optimize this using a broadcast channel
                     let built_msg = Message::text(msg.payload.to_string());
-                    for client in room.clients() {
-                        let Some(client) = ctx_read.clients.get(client) else {
-                            anyhow::bail!("client not found");
-                        };
-
-                        client.send(built_msg.clone()).await?;
+                    
+                    match ctx_read.clients.get(&msg.client_id) {
```
dyc3, line 627: This needs to check if `msg.client_id` is `Some`. That way we don't accidentally broadcast a message that was only intended for one client.

crates/ott-balancer-bin/src/balancer.rs
```diff
@@ -622,17 +622,23 @@ pub async fn dispatch_monolith_message(
...
+                        Some(client) => {
+                            client.send(built_msg.clone()).await?;
```
dyc3, line 629: No need to clone here.

dyc3 (changes requested):

client/src/locales/en.ts
dyc3, line 1: I'm not sure how these changes got in here, but this should revert it:
```bash
git restore --source=master client/src/locales/en.ts
```

dyc3 (changes requested):

PR #1116, opened 2023-10-24
fix join race again

### LINKED ISSUE #1101: Room Load/Client Join Race Condition
### Current Behavior

When a room is loaded on a monolith and a client attempts to join that room at the exact same time, a race condition occurs, and can result in the same room getting loaded on a different monolith.

### Expected Behavior

If joining a room causes another instance to be created, the new room needs to be unloaded and the clients sent to the existing room.

### Steps To Reproduce

1. Get a monolith to load a room
2. Join that room through the balancer

Timing is left as an exercise to the reader.

### Environment

### Anything else?

_No response_

dyc3:
- add load_notifiers
- possibly make it work better?
- keep load notifiers around after the room has loaded to prevent rugpulling state
- fix rare join race in multithreaded runtime
- adjust iterations for `route_ws_to_correct_monolith_race`

closes #1101 again, hopefully for the last time

### DIFF
--- a/crates/harness-tests/src/routing.rs
+++ b/crates/harness-tests/src/routing.rs
@@ -127,7 +127,7 @@ async fn route_ws_to_correct_monolith_race(ctx: &mut TestRunner) {
     m.show().await;
     tokio::time::sleep(Duration::from_millis(200)).await; // ensure that the monoliths are fully connected before sending the room load message
 
-    for i in 0..20 {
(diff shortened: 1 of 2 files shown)


dyc3:
I'm not entirely convinced this complexity is necessary considering what actual user traffic is going to look like. I think that attempting to join a room at the same time it's being loaded is going to be extremely rare.

PR #1124, opened 2023-11-02
add `region` field to balancer protocol

### LINKED ISSUE #1121: add region field to `M2BInit`
add a new field `region` to `M2BInit`, and add support in the Monolith

the monolith should derive the region from these places, in this order:
- `FLY_REGION` environment variable
- configuration, with default value "unknown"

dyc3:
closes #1121

### DIFF
--- a/crates/harness/src/monolith.rs
+++ b/crates/harness/src/monolith.rs
@@ -85,7 +85,10 @@ impl Monolith {
                 loop {
                     let (stream, _) = _listener.accept().await.unwrap();
                     let mut ws = tokio_tungstenite::accept_async(stream).await.unwrap();
-                    let init = M2BInit { port: http_port };
+                    let init = M2BInit {
+                        port: http_port,
+                        region: "unknown".into(),
+                    };
                     let msg = serde_json::to_string(&MsgM2B::Init(init)).unwrap();
                     ws.send(Message::Text(msg)).await.unwrap();
                     state.lock().unwrap().connected = true;
--- a/crates/ott-balancer-protocol/src/monolith.rs
+++ b/crates/ott-balancer-protocol/src/monolith.rs
@@ -93,6 +93,7 @@ pub enum MsgM2B {
 pub struct M2BInit {
     /// The port that the monolith is listening for HTTP requests on.
     pub port: u16,
+    pub region: String,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
--- a/server/balancer.ts
+++ b/server/balancer.ts
@@ -75,6 +75,7 @@ class BalancerManager {
 			type: "init",
 			payload: {
 				port: conf.get("port"),
+				region: conf.get("balancing.region"),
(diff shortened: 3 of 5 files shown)


crates/harness/src/monolith.rs
```diff
@@ -85,7 +85,10 @@ impl Monolith {
                 loop {
                     let (stream, _) = _listener.accept().await.unwrap();
                     let mut ws = tokio_tungstenite::accept_async(stream).await.unwrap();
-                    let init = M2BInit { port: http_port };
+                    let init = M2BInit {
+                        port: http_port,
+                        region: "unknown".into(),
```
cjrkoa, line 90: what does .into() do?

crates/ott-balancer-protocol/src/monolith.rs
```diff
@@ -93,6 +93,7 @@ pub enum MsgM2B {
 pub struct M2BInit {
     /// The port that the monolith is listening for HTTP requests on.
     pub port: u16,
+    pub region: String,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
```
cjrkoa, line 99: line 99?

server/balancer.ts
cjrkoa, line 1: This file won't let me comment on individual lines so:
wss.on()? 
this.emit() - emits to what?

crates/harness/src/monolith.rs
dyc3: It comes from the `Into` trait. https://doc.rust-lang.org/std/convert/trait.Into.html

You can read more in depth about `From`/`Into` here: https://practice.rs/type-conversions/from-into.html Specifically, this `into()` converts a `&'static str` into a `String`.

server/ott-config.ts
```diff
@@ -513,6 +519,11 @@ function postProcessConfig(): void {
 	if (conf.get("mail.enabled")) {
 		validateMail();
```
cjrkoa, line 520: what does mail mean in this context?

cjrkoa (approved):

server/balancer.ts
dyc3: wss is a `WebSocket.Server` from the `ws` node js package.

`on(event, handler)` runs `handler` whenever `event` is emitted by the object.

`this` in that context is `BalancerManager`, which has an event bus that emits events from `BalancerManagerEvemts` (which ive just realized I've typod)

server/ott-config.ts
dyc3: email. used for password recovery emails.

crates/ott-balancer-protocol/src/monolith.rs
dyc3: `derive` invokes derive macros. They are macros that automatically implement traits for you. For example, `#[derive(Clone)]` will automatically implement `Clone`, and allow you to call `clone()` on the type that it's implemented for.

PR #1132, opened 2023-11-06
add a script to automate bumping the version number

### LINKED ISSUE #1122: script to set version fields in all `package.json` and `cargo.toml` files

dyc3:
closes #1122

### DIFF
--- a/scripts/set-version.sh
+++ b/scripts/set-version.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+set -exo pipefail
+
+cd "$(dirname "$0")/.." || exit 1
+
+newVersion="$1"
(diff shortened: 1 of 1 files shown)


scripts/set-version.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+set -exo pipefail
```
cjrkoa, line 3: I think this line deals with pipeline failure but what does this line do?

dyc3: It's a one liner that does a couple things. See: https://gist.github.com/mohanpedala/1e2ff5661761d3abd0385e8223e16425

Issue #1129, opened 2023-11-03
harness: Generalize emulated monoliths so that behavior can be customized


dyc3:
### Is your feature request related to a problem? Please describe.

Sometimes it's useful for the emulated monoliths to have certain behaviors, for example: sending the init message as soon as it receives a connection.

However, some test cases require that emulated monoliths opt out of this automated behavior.

### Describe the solution you'd like

There should be a `Behavior` trait that `Monolith` can use to opt in to certain behaviors. Behaviors shouldn't keep track of monolith state so that it remains accessible in tests.

```rust
trait Behavior {
    fn on_connect(conn, state) {}
    fn on_disconnect(conn, state) {}
    fn on_message(conn, state, msg) {}
}
```

and we could have some auto impls for tuples of behaviors so that you can use multiple behaviors in one monolith.

```rust
impl<B1, B2> Behavior for (B1, B2) where B1: Behavior, B2: Behavior {
    fn foo() {
        B1::foo();
        B2::foo();
    }
}

impl<B1, B2, B3> Behavior for (B1, B2, B3) where B1: Behavior, B2: Behavior, B3: Behavior {
    fn foo() {
        B1::foo();
        B2::foo();
        B3::foo();
    }
}

...
```

a macro would probably be good to use for that.

Issue #1127, opened 2023-11-02
harness: add region to emulated monoliths


dyc3:
Add it to `MonolithBuilder` builder api, and the `MonolithState`.

cjrkoa:
I'd like to pick up this issue

PR #1133, opened 2023-11-07
add builder api for emulated monoliths

dyc3:
This will make it easier to work with emulated monoliths after #1127 and #1129 are done.

cc @cjrkoa

### DIFF
--- a/crates/harness-tests/src/routing.rs
+++ b/crates/harness-tests/src/routing.rs
@@ -1,6 +1,6 @@
 use std::time::Duration;
 
-use harness::{Client, MockRespParts, Monolith, TestRunner};
+use harness::{Client, MockRespParts, Monolith, MonolithBuilder, TestRunner};
 use test_context::test_context;
 
 #[test_context(TestRunner)]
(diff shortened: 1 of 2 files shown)


crates/harness-tests/src/routing.rs
```diff
@@ -39,20 +41,21 @@ async fn route_http_to_correct_monolith(ctx: &mut TestRunner) {
 #[test_context(TestRunner)]
 #[tokio::test]
 async fn route_http_to_any_monolith_once(ctx: &mut TestRunner) {
-    let mut m1 = Monolith::new(ctx).await.unwrap();
+    let mut m1 = MonolithBuilder::new()
+        .add_mock_http_json(
+            "/api/user",
+            MockRespParts::default(),
```
cjrkoa, line 47: line 47?

dyc3: see: https://doc.rust-lang.org/std/default/trait.Default.html

cjrkoa (approved):

PR #1134, opened 2023-11-08
add emulated monolith behaviors

### LINKED ISSUE #1129: harness: Generalize emulated monoliths so that behavior can be customized

dyc3:
- prototype emulated monolith behaviors
- implement Behavior for tuples of behaviors
- add BehaviorTrackClients and BehaviorLoadRooms

closes #1129


crates/harness/src/monolith.rs
```diff
@@ -45,7 +49,7 @@ pub struct Monolith {
 /// This is necessary because the monolith is split into two tasks: one for the websocket
 /// connection and one for the HTTP mock server.
 #[derive(Debug, Default)]
-pub(crate) struct MonolithState {
```
cjrkoa, line 48: I know this line was removed but what's the difference between pub(crate) struct and pub struct?

crates/harness/src/monolith.rs
```diff
@@ -58,6 +62,13 @@ pub(crate) struct MonolithState {
 
 impl Monolith {
     pub async fn new(ctx: &TestRunner) -> anyhow::Result<Self> {
+        Self::with_behavior(ctx, Box::new(BehaviorManual)).await
+    }
+
+    pub async fn with_behavior(
+        ctx: &TestRunner,
+        behavior: Box<dyn Behavior + Send + 'static>,
```
cjrkoa, line 70: I looked at the docs for dyn, I'm kinda confused. Why does this need to be dynamically dispatched?

crates/harness/src/monolith.rs
```diff
@@ -105,20 +116,20 @@ impl Monolith {
...
-                                                MsgB2M::Join(join) => {
-                                                    state.clients.insert(join.client);
-                                                },
-                                                MsgB2M::Leave(leave) => {
-                                                    state.clients.remove(&leave.client);
-                                                },
-                                                _ => {},
-                                            }
+                                        let to_send = {
+                                            let mut state = state.lock().unwrap();
```
cjrkoa, line 120: This only occurred to me now but I see .lock().unwrap() a lot. Why?

crates/harness/src/monolith.rs
```diff
@@ -391,7 +426,10 @@ impl MonolithBuilder {
     }
 
     pub async fn build(self, ctx: &TestRunner) -> Monolith {
-        let mut monolith = Monolith::new(ctx).await.unwrap();
+        let mut monolith =
+            Monolith::with_behavior(ctx, self.behavior.unwrap_or(Box::new(BehaviorManual)))
```
cjrkoa, line 430: Why do pointers need to be in a box?

crates/harness/src/monolith/behavior.rs
```diff
@@ -0,0 +1,172 @@
+use super::*;
+
+/// Defines how an emulated monolith should behave in response to messages from the balancer.
+pub trait Behavior: Send {
+    /// Processes the given message from the balancer, and returns messages to send back to the balancer.
+    fn on_msg(&self, msg: &MsgB2M, state: &mut MonolithState) -> Vec<MsgM2B>;
+}
+
+macro_rules! impl_behavior_tuple {
+    ($($ty:ident),*) => {
```
cjrkoa, line 10: $($ty:ident) ?????????

cjrkoa (approved):

crates/harness/src/monolith.rs
dyc3: `pub(crate)` makes it visible within the entire crate. `pub` makes it visible to external crates.

crates/harness/src/monolith.rs
dyc3: `state` is a `Arc<Mutex<T>>`. The `Arc` pointer gets implicitly dereferenced, but in order to safely access the contents of a Mutex, you have to lock it first. The result of that is you get a `MutexGuard` that automatically releases the lock when it's dropped.

crates/harness/src/monolith.rs
dyc3: A `Box` is just a heap allocated pointer. You have to do this because `dyn Trait` is not sized, and the compiler has to know how big each type is in order to allocate memory for it.

crates/harness/src/monolith.rs
dyc3: This avoids needing to have Monolith be a generic, which was causing other problems for me. I don't remember the specific reasons I did it like this.

crates/harness/src/monolith/behavior.rs
dyc3: This is macro syntax. `$ty:ident` means that the token it's looking for when parsing should be an identifier, and the `,*` makes it so that you can specify multiple separated by commas.

PR #1138, opened 2023-11-12
Add Region Configuration

### LINKED ISSUE #1126: balancer: add region to config
Needs to be able to pick up the `FLY_REGION` environment variable.

code link: https://github.com/dyc3/opentogethertube/blob/master/crates/ott-balancer-bin/src/config.rs
figment docs: https://docs.rs/figment/latest/figment/

cjrkoa:
Fixes #1126
In Progress

### DIFF
--- a/crates/ott-balancer-bin/src/config.rs
+++ b/crates/ott-balancer-bin/src/config.rs
@@ -56,6 +56,31 @@ impl BalancerConfig {
         // SAFETY: get is never called before CONFIG is initialized.
         unsafe { CONFIG.as_ref().expect("config not initialized") }
     }
+
+    pub fn set() -> &'static Self { 
+        figment::providers::Jail::expect_with(|jail| {
+            jail.create_file("Fly.toml", r#"
+                name = "Just a TOML App!"
+                count = 100
+            "#)?;
+        
+            jail.create_file("Fly.json", r#"
+                {
+                    "name": "Just a JSON App",
+                    "authors": ["figment", "developers"]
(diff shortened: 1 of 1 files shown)


crates/ott-balancer-bin/src/config.rs
```diff
@@ -56,6 +56,31 @@ impl BalancerConfig {
...
+                }
+            "#)?;
+        
+            jail.set_env("FLY_REGION", 250);
+        
+            // Sources are read _eagerly_: sources are read as soon as they are
+            // merged/joined into a figment.
+            let figment = figment::Figment::new()
+                .merge(figment::providers::Toml::file("Fly.toml"))
+                .merge(figment::providers::Env::prefixed("FLY_"))
```
dyc3, line 80: We can't accept all `FLY_` environment variables because it opens us up to configuration collision -- Fly could add a new environment var that overlaps with a config option, and that would be bad,

crates/ott-balancer-bin/src/config.rs
```diff
@@ -16,13 +16,15 @@ pub struct BalancerConfig {
...
     pub discovery: DiscoveryConfig,
+    pub region: String,
 }
 
 impl Default for BalancerConfig {
     fn default() -> Self {
         Self {
             port: 8081,
             discovery: DiscoveryConfig::default(),
+            region: "".to_string(), //TODO: Fix this
```
dyc3, line 27: default should be `unknown`

PR #1141, opened 2023-11-21
harness: add `TestRunner::restart_balancer()`

### LINKED ISSUE #1140: add function to restart balancer to harness
For some tests, we need to be able to restart the balancer so that we can assert how it recovers from crashes or on fresh start

dyc3:
closes #1140

### DIFF
--- a/crates/harness/src/test_runner.rs
+++ b/crates/harness/src/test_runner.rs
@@ -8,8 +8,7 @@ use tracing::warn;
 use crate::util::random_unused_port;
 
 pub struct TestRunner {
-    /// The port that the balancer is listening on for client connections.
-    pub port: u16,
+    spawn_options: BalancerSpawnOptions,
     pub(crate) child: Child,
 
     pub(crate) monolith_add_tx: tokio::sync::mpsc::Sender<SocketAddr>,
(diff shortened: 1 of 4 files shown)


crates/harness/src/test_runner.rs
```diff
@@ -8,8 +8,7 @@ use tracing::warn;
 use crate::util::random_unused_port;
 
 pub struct TestRunner {
-    /// The port that the balancer is listening on for client connections.
-    pub port: u16,
+    spawn_options: BalancerSpawnOptions,
```
cjrkoa, line 11: what does this do?

crates/harness/src/test_runner.rs
```diff
@@ -44,10 +43,10 @@ impl AsyncTestContext for TestRunner {
                 "--log-level",
                 "debug",
             ])
-            .env("BALANCER_PORT", format!("{}", port))
+            .env("BALANCER_PORT", format!("{}", opts.port))
```
cjrkoa, line 46: opts.port vs port?

cjrkoa (approved):

crates/harness/src/test_runner.rs
dyc3: BalancerSpawnOptions is used to hold the parameters used to start the balancer. They are held here to so they can be reused when restarting the balancer.

PR #1149, opened 2023-12-03
Add region to emulated monolith

### LINKED ISSUE #1127: harness: add region to emulated monoliths

cjrkoa:
Fixes #1127


crates/harness/src/monolith.rs
```diff
@@ -416,6 +421,7 @@ pub struct MockRequest {
 pub struct MonolithBuilder {
     response_mocks: HashMap<String, (MockRespParts, Bytes)>,
     behavior: Option<Box<dyn Behavior + Send + 'static>>,
+    region: String,
```
dyc3, line 424: I'd like the default for region to be `unknown` instead of an empty string. I would recommend just doing this in `new()`.

crates/harness/src/monolith.rs
```diff
@@ -460,4 +466,9 @@ impl MonolithBuilder {
         self.behavior = Some(Box::new(behavior));
         self
     }
+
+    pub fn region(mut self, region: String) -> Self {
```
dyc3, line 470: Would prefer this not require an owned `String`.

dyc3 (changes requested):
`MonolithBuilder::build` doesn't set the region on the monolith. Also, no unit tests?

dyc3 (changes requested):

crates/harness/src/monolith.rs
```diff
@@ -431,6 +441,7 @@ impl MonolithBuilder {
                 .await
                 .unwrap();
         monolith.set_all_mock_http(self.response_mocks);
+        monolith.set_region("unknown");
```
dyc3, line 444: This needs to reference the value set in `MonolithBuilder`, not `unknown`.

dyc3 (changes requested):

crates/harness-tests/src/state.rs
```diff
@@ -48,3 +48,11 @@ async fn should_unload_duplicate_rooms_and_route_correctly(ctx: &TestRunner) {
...
         .expect("timed out waiting for client join");
 }
+
+#[test_context(TestRunner)]
+#[tokio::test]
+async fn emulated_region(ctx: &TestRunner) {
+    let mb: Monolith = MonolithBuilder::new().region("foo").build(ctx).await.into();
+
+    assert_eq!(mb.region(), "foo");
+}
```
dyc3, line 58: This test should go inside `harness` because it's testing the harness.

Also, now you don't have a test that asserts the default region is "unknown".

crates/harness/src/monolith.rs
```diff
@@ -460,4 +471,10 @@ impl MonolithBuilder {
         self.behavior = Some(Box::new(behavior));
         self
     }
+
+    pub fn region(mut self, region: &str) -> Self {
```
dyc3, line 475: Would prefer taking in `impl AsRef<str>` instead.
```suggestion
    pub fn region(mut self, region: impl AsRef<str>) -> Self {
```

dyc3 (changes requested):

dyc3 (approved):

PR #1151, opened 2023-12-03
add unicast room messaging

### LINKED ISSUE #1100: balancer: Implement unicast room message handling

cjrkoa:
Fixes #1100

This should be the last new branch. When I created this branch, it explicitly told me it was based off of dyc3/master. Idk what to do if this still has the extra commits (On GitHub now: it says this branch has 1 commit so I think I'm good).

### DIFF
--- a/crates/harness-tests/src/routing.rs
+++ b/crates/harness-tests/src/routing.rs
@@ -2,6 +2,7 @@ use std::time::Duration;
 
 use harness::{Client, MockRespParts, Monolith, MonolithBuilder, TestRunner};
 use ott_balancer_protocol::monolith::M2BRoomMsg;
+use serde_json::value::RawValue;
 use test_context::test_context;
 
 #[test_context(TestRunner)]
@@ -220,7 +221,7 @@ async fn unicast_messaging(ctx: &mut TestRunner) {
     m.send(M2BRoomMsg {
         room: "foo".into(),
         client_id: c_id,
-        payload: RawValue::from_string("{}".to_owned()).unwrap(),,
+        payload: RawValue::from_string("{}".to_owned()).unwrap(),
     })
     .await;
 
--- a/crates/harness-tests/src/routing.rs
+++ b/crates/harness-tests/src/routing.rs
@@ -226,4 +226,6 @@ async fn unicast_messaging(ctx: &mut TestRunner) {
         payload: serde_json::json!({}),
     })
     .await;
+
+    assert_ne!(c1.recv(), c2.recv());
 }
--- a/crates/harness-tests/src/routing.rs
+++ b/crates/harness-tests/src/routing.rs
@@ -228,5 +228,9 @@ async fn unicast_messaging(ctx: &mut TestRunner) {
     })
     .await;
 
-    assert!(m.clients().get(c_id.as_ref().unwrap()).expect("{}"));
+    let res1 = c1.recv().await.unwrap();
+    let res2 = c2.recv().await.unwrap();
+
+    assert!(res1.to_string() == "{}");
+    assert!(res2.is_empty());
 }
--- a/crates/harness-tests/src/routing.rs
+++ b/crates/harness-tests/src/routing.rs
@@ -229,10 +229,8 @@ async fn unicast_messaging(ctx: &mut TestRunner) {
     .await;
 
     let res = vec![c1.recv().await, c2.recv().await];
+    let oks: Vec<_> = res.iter().filter(|r| r.is_ok()).collect();
 
-    assert_eq!(
-        (matches!(res.get(0), Ok(_)) && matches!(res.get(1), Err(_)))
-            || (matches!(res.get(0), Err(_)) && matches!(res.get(1), Ok(_))),
-        true
-    );
+    assert_eq!(oks.len(), 1);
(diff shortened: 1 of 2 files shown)


dyc3 (dismissed):

crates/harness-tests/src/routing.rs
```diff
@@ -199,3 +200,29 @@ async fn monolith_double_load_room(ctx: &mut TestRunner) {
...
+    c2.join("foo").await;
+
+    m.wait_recv().await;
+
+    let c_id = m.clients().iter().next().copied();
+
+    m.send(M2BRoomMsg {
+        room: "foo".into(),
+        client_id: c_id,
+        payload: serde_json::json!({}),
```
dyc3, line 223: This will fix the compile error
```suggestion
        payload: RawValue::from_string("{}".to_owned()).unwrap(),
```

crates/harness-tests/src/routing.rs
```diff
@@ -199,3 +200,29 @@ async fn monolith_double_load_room(ctx: &mut TestRunner) {
...
+    })
+    .await;
+
+    assert_ne!(c1.recv().await.into(), c2.recv().await.into());
```
dyc3, line 227: nit: a better assertion would be to check each client to see how many messages are pending.

crates/harness-tests/src/routing.rs
```diff
@@ -199,3 +200,29 @@ async fn monolith_double_load_room(ctx: &mut TestRunner) {
     let t = resp.text().await.expect("failed to read http response");
     assert_eq!(t, "{\"name\":\"foo\"}");
 }
+
+#[test_context(TestRunner)]
+#[tokio::test]
+async fn unicast_messaging(ctx: &mut TestRunner) {
+    let mut m = Monolith::new(ctx).await.unwrap();
```
dyc3, line 207: You'll need to opt in to the client tracking monolith behavior: `BehaviorTrackClients`. Use `MonolithBuilder`.

dyc3 (changes requested):
Build is still failing. I went ahead and fixed a couple of minor things for you, so pull before you make more changes

crates/harness-tests/src/routing.rs
```diff
@@ -225,5 +228,5 @@ async fn unicast_messaging(ctx: &mut TestRunner) {
     })
     .await;
 
-    assert_ne!(c1.recv().await.into(), c2.recv().await.into());
+    assert!(m.clients().get(c_id.as_ref().unwrap()).expect("{}"));
```
dyc3, line 231: `assert!()` takes a boolean value

dyc3 (changes requested):

crates/harness-tests/src/routing.rs
```diff
@@ -199,3 +201,36 @@ async fn monolith_double_load_room(ctx: &mut TestRunner) {
...
+
+    m.send(M2BRoomMsg {
+        room: "foo".into(),
+        client_id: c_id,
+        payload: RawValue::from_string("{}".to_owned()).unwrap(),
+    })
+    .await;
+
+    let res1 = c1.recv().await.unwrap();
+    let res2 = c2.recv().await.unwrap();
```
dyc3, line 232: `recv()` will time out if no message is received and return `Err` instead of `Ok`.

Instead of unwrapping them, you could put them in a Vec and make assertions on that. Something like this:
```rust
    let res = vec![c1.recv().await, c2.recv().await];
```
Then you just check to see if only one of the results in the list is `Ok` and matches the message.

crates/harness-tests/src/routing.rs
```diff
@@ -199,3 +201,36 @@ async fn monolith_double_load_room(ctx: &mut TestRunner) {
...
+
+    assert!(res1.to_string() == "{}");
```
dyc3, line 234: nit: prefer `assert_eq!` when asserting equality for better error messages

dyc3 (changes requested):
It would probably help if you run `cargo test` before pushing.

cjrkoa:
> It would probably help if you run `cargo test` before pushing.

I did run cargo test before pushing and it said the test passed

crates/harness-tests/src/routing.rs
```diff
@@ -199,3 +201,36 @@ async fn monolith_double_load_room(ctx: &mut TestRunner) {
...
+    m.show().await;
+
+    let mut c1 = Client::new(ctx).unwrap();
+    c1.join("foo").await;
+    let mut c2 = Client::new(ctx).unwrap();
+    c2.join("foo").await;
+
+    m.wait_recv().await;
+
+    let c_id = m.clients().iter().next().copied();
```
dyc3, line 222: HashSet does not maintain the order of items that are put in it, so this is not guarenteed to always get the client id of `c1`.

dyc3 (approved):

PR #1189, opened 2023-12-21
initial attempt to fix title > 255 char

Victor-M-Giraldo:
Not sure if this will work.

1. Still need to add form validator on frontend.
2. Took a look at pre-existing unit tests and not too sure how to write one for this. Will continue to look into it.


dyc3:
I took a quick took, turns out I don't have unit tests for the patch endpoint already, but I do have one for the create endpoint here: https://github.com/dyc3/opentogethertube/blob/67023d553e22ce85fb0bbd758d0d20cf5a7c8d87/server/tests/unit/api/room.spec.ts#L139

since you can include a title when creating a room too, you'll need to check the title in there too

Victor-M-Giraldo:
> I took a quick took, turns out I don't have unit tests for the patch endpoint already, but I do have one for the create endpoint here:
> 
> https://github.com/dyc3/opentogethertube/blob/67023d553e22ce85fb0bbd758d0d20cf5a7c8d87/server/tests/unit/api/room.spec.ts#L139
> 
> since you can include a title when creating a room too, you'll need to check the title in there too

Currently the change I made kind of works on my local copy? When I try to edit the title to be > 255 chars a 401 response is sent, but when you try to make any edits afterwards nothing works and the room settings cannot be loaded.

What are the differences between createRoom & generateRoom? In createRoom the title property does not exist. Also, if I use the same code I used for the patch request I get an error where the field is not recognized as a property. 

I figured I would need to mess with createRoom, but that seems to not be the case.

dyc3:
Yes, `createRoom` is the function that creates a room.

Making a HTTP request: `POST /api/room/create` calls `createRoom`, which creates a named room. Making a HTTP request: `POST /api/room/generate` calls `generate`, which creates a room with a randomly generated UUID name. 

Looks like you'll need to add `title?: string` to `OttApiRequestRoomCreate` to get rid of that error. I did not exactly make that type comprehensive.

> but when you try to make any edits afterwards nothing works and the room settings cannot be loaded.

I'll try it on my machine.

server/api/room.ts
```diff
@@ -237,6 +238,16 @@ const patchRoom: RequestHandler = async (req, res) => {
...
 	}
 
+	if (req.body.title && req.body.title.length > 255) {
+		res.status(401).json({
+			success: false,
+			error: {
+				field: "title",
+				message: "Room title must be 255 characters or less"
+			}
+		});
```
dyc3, line 248: This is the problem, you're responding to the request without returning, so execution continues and results in OTT trying to respond to the request twice, and you'll get a `ERR_HTTP_HEADERS_SENT` error and it'll shut down. 

You aren't meant to directly respond with an error like this because of how error prone it is. You should be throwing a `BadApiArgumentException`, like the other validators above. The error handler will automatically respond with the correct json and HTTP status.

dyc3 (changes requested):

server/api/room.ts
Victor-M-Giraldo: That's what I figured as you did similar validation on the name in createRoom. I will make those changes.

dyc3:
Make sure all the required CI is passing before you mark it as ready to review. `yarn lint` will fix formatting errors and some lints automatically, `yarn test` will run tests.  You might want to turn on format on save in vscode.

Victor-M-Giraldo:
> Make sure all the required CI is passing before you mark it as ready to review. `yarn lint` will fix formatting errors and some lints automatically, `yarn test` will run tests. You might want to turn on format on save in vscode.

Got it.

Victor-M-Giraldo:
I did run yarn test & yarn lint, as well as turning on format on save. I'm not sure why the tests are failing, however. The changes I made seem to be working though.

dyc3:
CI says tests pass, but there's formatting errors. You can click details on the failing job to see the logs. But tbh `yarn lint` should have fixed those things.

Victor-M-Giraldo:
It gives me 16 warnings, 0 errors. It's giving me those warnings in files I haven't touched at all yet though. Any idea why?

Also, two things.

I see that rust build failed to run and that is required to merge. Not sure why.

Additionally, when I run the test suite some test suites fail to run. Jest says it's finding 6 handles keeping it from exiting.

dyc3:
`yarn lint` will modify files for formatting without emitting errors, so check if you have any uncommitted changes.

>  It's giving me those warnings in files I haven't touched 
Don't worry about those. Haven't fixed them yet.

The rust tests are a bit flaky right now, so don't worry about it.

The jest open handle thing is normal, it doesn't affect the test results.

Victor-M-Giraldo:
Stupid mistake. Was on the wrong branch. Should hopefully be good now.

dyc3 (approved):
Diff looks good. Do you want to add unit tests for this? Totally optional, I'll add them if you don't.

Victor-M-Giraldo:
> Diff looks good. Do you want to add unit tests for this? Totally optional, I'll add them if you don't.

I want to try and give it a shot.

Victor-M-Giraldo:
Added the unit test for the create endpoint. Ran yarn lint & test.

Not sure if yarn lint did anything though.

Besides that, a few questions. 

In room.spec.ts where you describe the tests for the create endpoint, what do getSessionInfoSpy and validateSpy do?

What purpose do the tokens serve, are they a type of identifier?

If I am creating unit tests for the other endpoint, would I need to declare the app variable again and include the beforeAll, afterAll and afterEach again?

Also, in common/models are those files similar to mongoose schemas? Basically, are they how the database tables are set up?

server/tests/unit/api/room.spec.ts
```diff
@@ -225,6 +225,13 @@ describe("Room API", () => {
 					isTemporary: true,
 				},
 			],
+			[
+				{ arg: "title", reason: "not allowed (too long, must be at most 255 characters)" },
+				{
+					name: "abababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababab",
```
dyc3, line 231: The field you should be testing for is `title`, not `name`. It's still required to provide a `name` when creating a room, so set it to something valid like `foo`.

Room name is different from the room title. Room name is used like a unique identifier for the room, and appears in URLs.

dyc3:
> what do getSessionInfoSpy and validateSpy do?
Those are [jest mock functions](https://jestjs.io/docs/mock-function-api). They are mocked because calling the real versions of those functions requires redis, and I deliberately don't want to require redis to run unit tests because that would ultimately reduce test isolation.

> What purpose do the tokens serve?
The auth tokens are used to keep track of sessions. Sessions keep track of what account they are logged in as, or what username they currently have.

> If I am creating unit tests for the other endpoint, would I need to declare the app variable again and include the beforeAll, afterAll and afterEach again?

Inside the `describe("Room API", () => {` block, you don't need to define `app` again. It's already in scope. Inside of that block, you would need to add a new `describe`. Something like this:

```ts
describe("POST /api/room/:name", () => {
   beforeAll(() => {
       // create a temporary room to test against
    });

    afterAll(() => {
        // unload all rooms
    });

    it.each([
        [expectedError, requestBody],
    ])("should fail to modify room for validation errors: %s", async (error, body) => {
        // test code here
    });
});
```
No need to add a bunch of cases to get full coverage of the endpoint, just write enough to cover the changes in this PR.

The stuff in `common/models` is basically just type definitions shared across the server and client (its a poor name choice, haven't bothered to change it). Not really related to db models. Those are handled with sequelize, see `server/models`.

dyc3 (changes requested):

server/tests/unit/api/room.spec.ts
Victor-M-Giraldo: Made those changes. Should be good now I think.

server/tests/unit/api/room.spec.ts
```diff
@@ -1,4 +1,4 @@
-import _ from "lodash";
+import _, { before } from "lodash";
```
dyc3, line 1: I think this is a stray import

dyc3 (changes requested):
Looks good. just a minor thing

server/tests/unit/api/room.spec.ts
Victor-M-Giraldo: Should be good.

dyc3 (approved):

server/tests/unit/api/room.spec.ts
```diff
@@ -9,6 +9,7 @@ import { Room as RoomModel, User as UserModel } from "../../../models";
 import usermanager from "../../../usermanager";
 import { OttApiRequestRoomCreate } from "common/models/rest-api";
 import { conf } from "../../../../server/ott-config";
+import { error } from "console";
```
dyc3, line 12: I don't think you meant to import this

dyc3 (approved):

PR #1202, opened 2024-01-06
balancer: fix infinite loop in client connection handling

### LINKED ISSUE #1187: Balancer stops responding to all HTTP requests when Monolith stops
### Current Behavior

When a monolith restarts, the balancer becomes completely unresponsive.

### Expected Behavior

1. Balancer should disconnect all clients connected to the monolith
2. Balancer should continue responding to HTTP requests, including `GET /api/status`

### Steps To Reproduce

1. Set up 2 monolith, 1 balancer
2. Create a room
3. Stop the monolith that has the room
```
fly machines stop --app ott-staging --select
```
4. See that Balancer becomes unresponsive, no longer responds to `GET /api/status`

### Environment

### Anything else?

Found when manually testing on the staging deployment.

This is most likely the result of a deadlock.

dyc3:
fixes #1187

### DIFF
--- a/crates/ott-balancer/src/client.rs
+++ b/crates/ott-balancer/src/client.rs
@@ -183,36 +183,38 @@ pub async fn client_entry<'r>(
 
     loop {
         tokio::select! {
-            msg = client_link.outbound_recv() => {
-                if let Ok(SocketMessage::Message(msg)) = msg {
+            Ok(msg) = client_link.outbound_recv() => {
+                if let SocketMessage::Message(msg) = msg {
                     if let Err(err) = stream.send(msg).await {
                         error!("Error sending ws message to client: {:?}", err);
                         break;
                     }
                 } else {
-                    continue;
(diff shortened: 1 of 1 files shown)


crates/ott-balancer/src/client.rs
```diff
@@ -183,36 +183,38 @@ pub async fn client_entry<'r>(
...
                 } else {
-                    continue;
+                    info!("Client outbound stream ended");
+                    break;
                 }
             }
 
-            msg = stream.next() => {
-                if let Some(Ok(msg)) = msg {
+            Some(msg) = stream.next() => {
```
Victor-M-Giraldo, line 198: Is this using pattern matching?

Victor-M-Giraldo (approved):

crates/ott-balancer/src/client.rs
dyc3: Yes, see the docs for `tokio::select!`

PR #1212, opened 2024-01-11
balancer: fix clients not being removed from balancer context on disconnect

dyc3:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved client management by enabling the removal of clients from the active list.

### DIFF
--- a/crates/ott-balancer/src/balancer.rs
+++ b/crates/ott-balancer/src/balancer.rs
@@ -212,6 +212,7 @@ impl BalancerContext {
         let monolith = self.find_monolith_mut(client_id)?;
         monolith.remove_client(client_id);
         monolith.send(B2MLeave { client: client_id }).await?;
+        self.clients.remove(&client_id);
 
         Ok(())
     }
--- a/crates/ott-balancer/src/balancer.rs
+++ b/crates/ott-balancer/src/balancer.rs
@@ -696,3 +696,87 @@ pub async fn dispatch_monolith_message(
 
     Ok(())
 }
+
+#[cfg(test)]
+mod test {
+    use std::net::Ipv4Addr;
+
+    use crate::discovery::{HostOrIp, MonolithConnectionConfig};
+
+    use super::*;
+
+    #[tokio::test]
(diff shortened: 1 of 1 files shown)


crates/ott-balancer/src/balancer.rs
```diff
@@ -696,3 +696,87 @@ pub async fn dispatch_monolith_message(
...
+
+    use crate::discovery::{HostOrIp, MonolithConnectionConfig};
+
+    use super::*;
+
+    #[tokio::test]
+    async fn test_clients_add_remove() {
+        // a bunch of setup
+        let room_name = RoomName::from("test");
+        let mut ctx = BalancerContext::new();
```
Victor-M-Giraldo, line 712: `BalancerContext` is simply a struct that describes the state of the balancer, including quantity of rooms per monolith, how many instances of a monolith per region, etc. correct?

crates/ott-balancer/src/balancer.rs
```diff
@@ -696,3 +696,87 @@ pub async fn dispatch_monolith_message(
...
+            NewMonolith {
+                id: monolith_id,
+                region: "unknown".into(),
+                config: MonolithConnectionConfig {
+                    host: HostOrIp::Ip(Ipv4Addr::LOCALHOST.into()),
+                    port: 3002,
+                },
+                proxy_port: 3000,
+            },
+            monolith_outbound_tx,
```
Victor-M-Giraldo, line 728: Why are these defined in a different scope to that of the other values? Is it because the `new()` constructor takes them as arguments?

Victor-M-Giraldo (approved):

crates/ott-balancer/src/balancer.rs
dyc3: yeah

crates/ott-balancer/src/balancer.rs
dyc3: The channels need to be defined at the function level context so they stay alive for the entire unit test. Otherwise the channels get disposed immediately and the test will fail.

PR #1216, opened 2024-01-11
use arc for broadcast channels

### LINKED ISSUE #1199: Use `Arc` for room broadcast channels
Broadcasting to all clients in a room requires each receiver to `clone()` a new copy of the message, causing unnecessary memory allocations. Wrapping the messages in an `Arc` should allow us to minimize the amount of memory being cloned.

Victor-M-Giraldo:
closes #1199

## Summary by CodeRabbit

- **Refactor**
	- Enhanced the transaction broadcasting mechanism for improved stability and performance.

### DIFF
--- a/crates/ott-balancer/src/monolith.rs
+++ b/crates/ott-balancer/src/monolith.rs
@@ -149,7 +149,7 @@ pub struct Room {
     metadata: Option<RoomMetadata>,
 
     /// The Sender used to broadcast to all clients in this room.
-    broadcast_tx: tokio::sync::broadcast::Sender<SocketMessage>,
+    broadcast_tx: Arc<tokio::sync::broadcast::Sender<SocketMessage>>,
 }
 
 impl Room {
@@ -159,7 +159,7 @@ impl Room {
             name,
             clients: Vec::new(),
             metadata: None,
-            broadcast_tx,
+            broadcast_tx: Arc::new(broadcast_tx),
         }
     }
 
--- a/crates/ott-balancer/src/monolith.rs
+++ b/crates/ott-balancer/src/monolith.rs
@@ -149,7 +149,7 @@ pub struct Room {
     metadata: Option<RoomMetadata>,
 
     /// The Sender used to broadcast to all clients in this room.
-    broadcast_tx: Arc<tokio::sync::broadcast::Sender<SocketMessage>>,
+    broadcast_tx: tokio::sync::broadcast::Sender<Arc<SocketMessage>>,
 }
 
 impl Room {
@@ -159,7 +159,7 @@ impl Room {
             name,
             clients: Vec::new(),
             metadata: None,
-            broadcast_tx: Arc::new(broadcast_tx),
+            broadcast_tx,
         }
     }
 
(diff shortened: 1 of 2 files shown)


Victor-M-Giraldo:
Do you know why the cypress tests fail? I took a look at the logs and it seems like they are failing to fetch the imported modules for the component tests, and it can't find a `.video button` for the e2e tests.

Victor-M-Giraldo:
Also, do you know if it would it be better to use `into()` to change the type of `broadcast_tx` as opposed to using `new()` or does it not matter too much?

Does one use more memory as opposed to the other?

crates/ott-balancer/src/monolith.rs
```diff
@@ -149,7 +149,7 @@ pub struct Room {
     metadata: Option<RoomMetadata>,
 
     /// The Sender used to broadcast to all clients in this room.
-    broadcast_tx: tokio::sync::broadcast::Sender<SocketMessage>,
+    broadcast_tx: Arc<tokio::sync::broadcast::Sender<SocketMessage>>,
```
dyc3, line 152: This is not what I'm looking for. The objective is to reduce the size of memory allocations for clones on messages that gets passed through the channel. For every receiver on this broadcast channel, the message _must_ be cloned for memory safety, as indicated [in the broadcast channel docs here](https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html).

What you've done here is effectively a no-op. It should be `Sender<Arc<SocketMessage>>`. `Arc` is short for an atomic reference counter. This seems like a good article if you want to learn more: https://reintech.io/blog/understanding-implementing-rust-arc-mutex

dyc3 (changes requested):

dyc3:
> Do you know why the cypress tests fail?

It's completely out of our control, and its really annoying. See https://github.com/cypress-io/cypress/issues/25913

crates/ott-balancer/src/monolith.rs
Victor-M-Giraldo: I have one more type mismatch I can't figure out how to fix. It's in `client.rs` in the `client_entry()` fn. 

```
loop {
        tokio::select! {
            Ok(msg) = client_link.outbound_recv() => {
                if let SocketMessage::Message(msg) = msg {
                    if let Err(err) = stream.send(msg).await {
                        error!("Error sending ws message to client: {:?}", err);
                        break;
                    }
                } else {
                    info!("Client outbound stream ended");
                    break;
                }
            }

            Some(msg) = stream.next() => {
                if let Ok(msg) = msg {
                    if let Err(err) = client_link.inbound_send(msg).await {
                        error!("Error sending client message to balancer: {:?}", err);
                        break;
                    }
                } else {
                    debug!("Client inbound websocket stream ended");
                    break;
                }
            }

            else => {
                debug!("Client websocket stream ended");
                break;
            }
        }
    }

    info!("ending client connection");

    Ok(())
}
```

In the `if let SocketMessage::Message(msg) = msg {` line how can I make `SocketMessage::Message(msg)` a `Arc<SocketMessage>`?

Here's the full error:
```
188 |                 if let SocketMessage::Message(msg) = msg {
    |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^   --- this expression has type `Arc<messages::SocketMessage>`
    |                        |
    |                        expected `Arc<SocketMessage>`, found `SocketMessage`
    |
    = note: expected struct `Arc<messages::SocketMessage>`
                 found enum `messages::SocketMessage`
```

dyc3: Make this a draft and push what you have so I can play with it

dyc3:
I took a look at it, and the problem is that the Stream wants owned values because it has to queue them internally before they get flushed to IO. I don't think this is feasible for us right now, so we can drop this task.

Victor-M-Giraldo:
No problem. Do you think you could elaborate a little on what you mean just for my own understanding?

From what I understand stream can't take a reference to a variable. It must have ownership of `msg` but what does this have to do with `Arc` and what we were trying to do?

Thanks.

dyc3:
We get `Arc<SocketMessage>` from the channel, but we need to turn it into a `Message` (from the tungstenite crate) in order to call `send`. This is impossible to do without cloning the inner value of the `Arc`, which is what we are trying to avoid, or using unsafe code, which we also want to avoid.

PR #1238, opened 2024-01-19
add package for visualizations

cjrkoa:
Having trouble getting Vue setup. Current error: No loader is configured for ".node" files: ../node_modules/fsevents/fsevents.node

closes dyc3/senior-design#227


balancer-visualization/balancer-visualization/.gitignore
dyc3, line 1: most of these are already ignored by the top level gitignore

balancer-visualization/balancer-visualization/yarn.lock
dyc3, line 1: remove this, the root level yarn.lock will be used instead when you add the package to the workspace

balancer-visualization/balancer-visualization/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "dependencies": {
+    "core-js": "^3.8.3",
+    "vue": "^3.2.13"
+  },
+  "devDependencies": {
+    "@babel/core": "^7.12.16",
+    "@babel/eslint-parser": "^7.12.16",
+    "@vue/cli-plugin-babel": "~5.0.0",
+    "@vue/cli-plugin-eslint": "~5.0.0",
+    "@vue/cli-service": "~5.0.0",
```
dyc3, line 19: You should be using vite like we already do for the client, not the cli service

dyc3 (changes requested):
- I'd prefer this package to go in `packages/debug` eg rename the package
- You'll need to add this package to the workspace in the root `package.json`
- Shouldn't we be using the grafana template as a base here?

PR #1243, opened 2024-01-20
abstract balancing logic 

### LINKED ISSUE #1195: balancer: abstract balancing logic into a trait

Currently, the primary logic that actually does load balancing is 2 functions on `BalancerContext`: 
- `select_monolith` - for selecting a monolith to load a new room on
- `random_monolith` - for routing requests that could go to any monolith

If we wanted to try new balancing algorithms, we would have to gut the current one and replace it, making it harder to switch back if it doesn't work (revert the commit, recompile, redeploy). If we had something hot swappable, it should be as easy to switch as a config change and restart. This would also allow for A/B testing of load balancing methods.

- Create a trait that can be used to switch between load balancing algorithms,
- It should attempt to deduplicate the code that prioritizes monoliths in the same region. As such, it shouldn't require implementers of the trait to also implement that logic.
- Ideally, eliminate the need for both `select_monolith` and `random_monolith`.

### Additional context

_No response_

Victor-M-Giraldo:
closes #1195

### DIFF
--- a/crates/ott-balancer/src/balancer.rs
+++ b/crates/ott-balancer/src/balancer.rs
@@ -369,29 +369,37 @@ impl BalancerContext {
             .ok_or(anyhow::anyhow!("monolith not found"))?;
         Ok(monolith)
     }
+
     /// When loading a room, call this to select the best monolith to load it on.
-    pub fn select_monolith(&self) -> anyhow::Result<&BalancerMonolith> {
+    pub fn filter_monoliths(&self) -> anyhow::Result<&BalancerMonolith> {
         fn cmp(x: &BalancerMonolith, y: &BalancerMonolith) -> std::cmp::Ordering {
             x.rooms().len().cmp(&y.rooms().len())
         }
 
+        // Retrieves a reference to a collection of MonolithIds specific to a region.
         let in_region = self
             .monoliths_by_region
             .get(BalancerConfig::get().region.as_str());
+        // Checks if Monoliths available for specified region?
         if let Some(in_region) = in_region {
             let selected = in_region
                 .iter()
                 .flat_map(|id| self.monoliths.get(id))
-                .min_by(|x, y| cmp(x, y));
-            if let Some(s) = selected {
-                return Ok(s);
-            }
+                .collect::<Vec<_>>(); // Turn iterators into vectors
+
+            // Don't need this anymore because no longer returning an option
+            // if let Some(s) = selected {
+            //     return Ok(s);
+            // }
         }
-        let selected = self.monoliths.values().min_by(|x, y| cmp(x, y));
-        match selected {
-            Some(s) => Ok(s),
-            None => anyhow::bail!("no monoliths available"),
-        }
+
+        // let selected = self.monolith_selection.select_monolith(monolith: Vec<&BalancerMonolith>);
+
+        // Don't need the pattern matching here as that's taken care of by select_monolith
+        // match selected {
+        //     Some(s) => Ok(s),
+        //     None => anyhow::bail!("no monoliths available"),
+        // }
     }
 
     pub fn random_monolith(&self) -> anyhow::Result<&BalancerMonolith> {
@@ -407,7 +415,7 @@ impl BalancerContext {
             }
         }
 
-        let selected = self
+        let selected = self 
             .monoliths
             .values()
             .choose(&mut rand::thread_rng())
--- a/crates/ott-balancer/src/balancer.rs
+++ b/crates/ott-balancer/src/balancer.rs
@@ -16,6 +16,8 @@ use crate::client::ClientLink;
 use crate::config::BalancerConfig;
 use crate::monolith::Room;
 use crate::room::RoomLocator;
+use crate::selection::MonolithRegistry;
+use crate::selection::MonolithSelection;
 use crate::{
     client::{BalancerClient, NewClient},
     messages::*,
@@ -179,19 +181,6 @@ pub struct BalancerContext {
     pub monoliths_by_region: HashMap<String, Vec<MonolithId>>,
     pub monolith_selection: Box<dyn MonolithSelection + Send + Sync + 'static>,
 }
-#[derive(Debug, Default)]
-pub struct MonolithRegistry;
-pub trait MonolithSelection: std::fmt::Debug {
-    fn select_monolith<'a>(
-        &'a self,
-        monolith: Vec<&'a BalancerMonolith>,
-    ) -> anyhow::Result<&BalancerMonolith>;
-
-    fn random_monolith<'a>(
-        &'a self,
-        monolith: Vec<&'a BalancerMonolith>,
-    ) -> anyhow::Result<&BalancerMonolith>;
-} 
 
 impl MonolithSelection for MonolithRegistry {
     fn select_monolith<'a>(
--- a/crates/ott-balancer/src/balancer.rs
+++ b/crates/ott-balancer/src/balancer.rs
@@ -171,7 +171,7 @@ impl BalancerLink {
     }
 }
 
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct BalancerContext {
     pub clients: HashMap<ClientId, BalancerClient>,
     pub monoliths: HashMap<MonolithId, BalancerMonolith>,
@@ -179,10 +179,9 @@ pub struct BalancerContext {
     pub monoliths_by_region: HashMap<String, Vec<MonolithId>>,
     pub monolith_selection: Box<dyn MonolithSelection + Send + Sync + 'static>,
 }
-// #[derive(Debug, Default)]
+#[derive(Debug, Default)]
 pub struct MonolithRegistry;
-
-pub trait MonolithSelection {
+pub trait MonolithSelection: std::fmt::Debug {
     fn select_monolith(
         &self,
         monolith: Vec<&BalancerMonolith>,
@@ -199,7 +198,7 @@ pub trait MonolithSelection {
 impl MonolithSelection for MonolithRegistry {
     fn select_monolith(
         &self,
-        monolith: Vec<&BalancerMonolith>,
+        _monolith: Vec<&BalancerMonolith>,
         ctx: &BalancerContext,
     ) -> anyhow::Result<&BalancerMonolith> {
         fn cmp(x: &BalancerMonolith, y: &BalancerMonolith) -> std::cmp::Ordering {
@@ -227,7 +226,7 @@ impl MonolithSelection for MonolithRegistry {
 
     fn random_monolith(
         &self,
-        monolith: Vec<&BalancerMonolith>,
+        _monolith: Vec<&BalancerMonolith>,
         ctx: &BalancerContext,
     ) -> anyhow::Result<&BalancerMonolith> {
         let in_region = ctx
@@ -248,6 +247,18 @@ impl MonolithSelection for MonolithRegistry {
             .choose(&mut rand::thread_rng())
             .ok_or(anyhow::anyhow!("no monoliths available"))?;
         Ok(selected)
+    }
+}
+
+impl Default for BalancerContext {
+    fn default() -> Self {
+        BalancerContext {
+            clients: HashMap::default(),
+            monoliths: HashMap::default(),
+            rooms_to_monoliths: HashMap::default(),
+            monoliths_by_region: HashMap::default(),
+            monolith_selection: Box::new(MonolithRegistry::default()),
+        }
     }
 }
 
--- a/crates/ott-balancer/src/balancer.rs
+++ b/crates/ott-balancer/src/balancer.rs
@@ -177,12 +177,12 @@ pub struct BalancerContext {
     pub monoliths: HashMap<MonolithId, BalancerMonolith>,
     pub rooms_to_monoliths: HashMap<RoomName, RoomLocator>,
     pub monoliths_by_region: HashMap<String, Vec<MonolithId>>,
-    pub monolith_selector: Box<dyn MonolithSelectors>,
+    pub monolith_selection: Box<dyn MonolithSelection + Send + Sync + 'static>,
 }
-
+// #[derive(Debug, Default)]
 pub struct MonolithRegistry;
 
-pub trait MonolithSelectors {
+pub trait MonolithSelection {
     fn select_monolith(
         &self,
         monolith: Vec<&BalancerMonolith>,
@@ -196,7 +196,7 @@ pub trait MonolithSelectors {
     ) -> anyhow::Result<&BalancerMonolith>;
 }
 
-impl MonolithSelectors for MonolithRegistry {
+impl MonolithSelection for MonolithRegistry {
     fn select_monolith(
         &self,
         monolith: Vec<&BalancerMonolith>,
--- a/crates/ott-balancer/src/balancer.rs
+++ b/crates/ott-balancer/src/balancer.rs
@@ -16,8 +16,7 @@ use crate::client::ClientLink;
 use crate::config::BalancerConfig;
 use crate::monolith::Room;
 use crate::room::RoomLocator;
-use crate::selection::MonolithRegistry;
-use crate::selection::MonolithSelection;
+use crate::selection::{MinRoomsSelector, MonolithSelection};
 use crate::{
     client::{BalancerClient, NewClient},
     messages::*,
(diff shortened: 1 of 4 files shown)


Victor-M-Giraldo:
Still needs work. Did you mean something like this though?

Victor-M-Giraldo:
Also, the is there a way for the new struct to just reference the monoliths defined in BalancerContext?

crates/ott-balancer/src/balancer.rs
```diff
@@ -179,6 +179,65 @@ pub struct BalancerContext {
     pub monoliths_by_region: HashMap<String, Vec<MonolithId>>,
 }
 
+pub struct MonolithRegistry {
+    pub monoliths_by_region: HashMap<String, Vec<MonolithId>>,
+    pub monoliths: HashMap<MonolithId, BalancerMonolith>,
+}
```
dyc3, line 185: Ideally, this would be a zero size type with no members. These maps should remain in `BalancerContext` and not cloned.

crates/ott-balancer/src/balancer.rs
```diff
@@ -179,6 +179,65 @@ pub struct BalancerContext {
...
+
+pub trait MonolithSelection {
+    
+    fn select_monolith(&self) -> anyhow::Result<&BalancerMonolith>;
+
+    fn random_monolith(&self) -> anyhow::Result<&BalancerMonolith>;
+}
```
dyc3, line 192: and then these functions should take something like a `Vec<&BalancerMonolith>` or `impl Iter<&BalancerMonolith>` as input so the `in_region` bits can be deduplicated

dyc3 (commented):
Kinda sorta. You should look at how the `Behavior` trait is used in `harness/src/monolith/behavior.rs`

crates/ott-balancer/src/balancer.rs
Victor-M-Giraldo: I made this change. Since `select_monolith` and `random_monolith` rely on `monoliths` and `monoliths_by_region` defined in `BalancerContext` for the load balancing logic, how do we make these available to the new struct without cloning them? 

That's why I asked that second question, because I knew we shouldn't clone those two members, but the function logic depends on them. I'm taking a look at `behavior.rs` right now.

crates/ott-balancer/src/balancer.rs
Victor-M-Giraldo: Just to clarify, `BalancerMonolith` refers to an instance of the Monolith & `BalancerContext` refers to the state the balancer is currently in, with the amount of clients, monoliths, rooms per monolith, etc. Is that right?

dyc3: Yes. 

`BalancerMonolith` directly corresponds to a websocket connection being maintained by the connection manager to the monolith.

crates/ott-balancer/src/balancer.rs
dyc3: >  without cloning them

By borrowing them, as references.

Victor-M-Giraldo:
Is this most recent commit any closer? Or would you like to keep the inputs to both functions only `&self` and `Vec<&BalancerMonolith>`.

dyc3:
Try to actually use this in `BalancerContext` to replace the code that's there, and when it compiles I'll be able to give you better feedback.

Victor-M-Giraldo:
> Try to actually integrate this (go and replace `select_monolith` and `random_monolith`), and when it compiles I'll be able to give you better feedback.

When we first discussed this, you said to do something like

```
struct BalancerContext {
    ....
    Box<dyn YourTrait> foo;
}
```

I'm having some trouble with that. Lots of compile errors, mostly regarding futures being sent between threads being unsafe & not implementing certain traits.

crates/ott-balancer/src/balancer.rs
```diff
@@ -177,11 +177,12 @@ pub struct BalancerContext {
     pub monoliths: HashMap<MonolithId, BalancerMonolith>,
     pub rooms_to_monoliths: HashMap<RoomName, RoomLocator>,
     pub monoliths_by_region: HashMap<String, Vec<MonolithId>>,
+    pub monolith_selector: Box<dyn MonolithSelectors>,
 }
 
 pub struct MonolithRegistry;
 
-pub trait MonolithSelection {
+pub trait MonolithSelectors {
```
dyc3, line 185: nit: make this singular

dyc3:
> wouldn't I need whatever type that is implementing my trait to also implement the traits that the compilers talking about?

yeah, that's exactly what the compiler is complaining about. realistically it'll be something like `Box<dyn MonolithSelector + Send>`

Victor-M-Giraldo:
>  it'll be something like `Box<dyn MonolithSelector + Send>`

Looks like it is both `Sync` and `Send`. 

From what I'm reading these are marker traits, so they have no associated methods.

Does including the `Send` & `Sync` traits just tell the compiler it's okay for instances to be sent and shared between threads?

I have to implement `Debug` and `Default` as well since `BalancerContext` does (I believe that's the reason). I thought it would be as simple as putting `#[derive(Debug, Default)]` before the new struct, but this doesn't work.

Victor-M-Giraldo:
```
pub trait MonolithSelection: std::fmt::Debug {

}
```

fixes the `Debug` error, the same does not work for `Default`

dyc3:
> Does including the `Send` & `Sync` traits just tell the compiler it's okay for instances to be sent and shared between threads?

That's the short version. The documentation for those goes into more detail.

crates/ott-balancer/src/balancer.rs
```diff
@@ -177,6 +177,78 @@ pub struct BalancerContext {
     pub monoliths: HashMap<MonolithId, BalancerMonolith>,
     pub rooms_to_monoliths: HashMap<RoomName, RoomLocator>,
     pub monoliths_by_region: HashMap<String, Vec<MonolithId>>,
+    pub monolith_selection: Box<dyn MonolithSelection + Send + Sync + 'static>,
+}
+// #[derive(Debug, Default)]
+pub struct MonolithRegistry;
```
dyc3, line 183: This derive should work fine

Victor-M-Giraldo: That's what I thought as well, but the compiler tells me those traits aren't being implemented.

dyc3: Played around with it a bit. The reason this doesn't work is because there is no `Default` for `Box`, and `Default` is derived on `BalancerContext`, so it's trying to derive a `Default` impl for a type it doesn't have. 

The easiest solution is probably to manually impl `Default` for `BalancerContext`

Victor-M-Giraldo: That looks like it works. 

Now I have some errors with lifetimes. I'm going to leave it here for today, but I would appreciate quite a bit more help tomorrow since lifetimes are a pretty novel concept to me, and frankly I don't understand them at all.

Thanks for your help today.

dyc3: Rust by example is a pretty good resource for this: https://doc.rust-lang.org/rust-by-example/scope/lifetime.html You might find the "Traits" and "Bounds" examples of particular interest.

crates/ott-balancer/src/balancer.rs
```diff
@@ -171,12 +171,95 @@ impl BalancerLink {
...
     pub monoliths_by_region: HashMap<String, Vec<MonolithId>>,
+    pub monolith_selection: Box<dyn MonolithSelection + Send + Sync + 'static>,
+}
+#[derive(Debug, Default)]
+pub struct MonolithRegistry;
+pub trait MonolithSelection: std::fmt::Debug {
+    fn select_monolith(
+        &self,
+        monolith: Vec<&BalancerMonolith>,
+        ctx: &BalancerContext,
```
dyc3, line 188: don't take `ctx` as an argument, that tightly couples the trait to `BalancerContext` and will make it harder to write unit tests

crates/ott-balancer/src/balancer.rs
```diff
@@ -171,12 +171,95 @@ impl BalancerLink {
...
     pub rooms_to_monoliths: HashMap<RoomName, RoomLocator>,
     pub monoliths_by_region: HashMap<String, Vec<MonolithId>>,
+    pub monolith_selection: Box<dyn MonolithSelection + Send + Sync + 'static>,
+}
+#[derive(Debug, Default)]
+pub struct MonolithRegistry;
+pub trait MonolithSelection: std::fmt::Debug {
+    fn select_monolith(
+        &self,
+        monolith: Vec<&BalancerMonolith>,
```
dyc3, line 187: You only need to operate on this list of `&BalancerMonolith`...

crates/ott-balancer/src/balancer.rs
```diff
@@ -171,12 +171,95 @@ impl BalancerLink {
...
+            .get(BalancerConfig::get().region.as_str());
+        if let Some(in_region) = in_region {
+            let selected = in_region
+                .iter()
+                .flat_map(|id| ctx.monoliths.get(id))
+                .min_by(|x, y| cmp(x, y));
+            if let Some(s) = selected {
+                return Ok(s);
+            }
+        }
```
dyc3, line 219: ... because this code should not be in the trait impl. The trait's purpose is not to do any filtering or preprocessing, its just to prioritize which monolith to use.

crates/ott-balancer/src/balancer.rs
```diff
@@ -171,12 +171,95 @@ impl BalancerLink {
...
+    pub monolith_selection: Box<dyn MonolithSelection + Send + Sync + 'static>,
+}
+#[derive(Debug, Default)]
+pub struct MonolithRegistry;
+pub trait MonolithSelection: std::fmt::Debug {
+    fn select_monolith(
+        &self,
+        monolith: Vec<&BalancerMonolith>,
+        ctx: &BalancerContext,
+    ) -> anyhow::Result<&BalancerMonolith>;
```
dyc3, line 189: The reason the compiler is complaining about lifetimes is because it can't automatically guarantee that the reference you are returning isn't something you fabricated in the function. The lifetime annotations tell the compiler that yes, the parameters and the returned object must live for the same amount of time. The compiler diagnostic gives a hint for what code changes need to be made if you look at the output in your terminal.

dyc3 (commented):
Played around with it a little more

Victor-M-Giraldo:
Still need to move this to a new module. Working on that now.

crates/ott-balancer/src/balancer.rs
```diff
@@ -171,12 +173,53 @@ impl BalancerLink {
...
+#[derive(Debug)]
 pub struct BalancerContext {
     pub clients: HashMap<ClientId, BalancerClient>,
     pub monoliths: HashMap<MonolithId, BalancerMonolith>,
     pub rooms_to_monoliths: HashMap<RoomName, RoomLocator>,
     pub monoliths_by_region: HashMap<String, Vec<MonolithId>>,
+    pub monolith_selection: Box<dyn MonolithSelection + Send + Sync + 'static>,
+}
+
+impl MonolithSelection for MonolithRegistry {
```
dyc3, line 185: You should keep the impl with the struct it's intended for. Move it to the new module.

crates/ott-balancer/src/selection.rs
```diff
@@ -0,0 +1,15 @@
+use crate::monolith::BalancerMonolith;
+
+#[derive(Debug, Default)]
+pub struct MonolithRegistry;
```
dyc3, line 4: nit: `MonolithRegistry` isn't a great name. It should describe the strategy it uses to balance load, which is by minimizing the rooms on each Monolith. So something like `MinRoomsSelector`.

dyc3 (commented):
This is looking good. You should start integrating this.

Ideally, you could create some unit tests for this too.

Victor-M-Giraldo:
I'm trying to implement this now. We agreed to change the name of `select_monolith` and `random_monolith` in the `BalancerContext` impl to something like `filter_monolith`. 

Since in each of these cases, the filtering logic is different, I can't just make this a new function, right?

For the implementation what I've been trying to do is something like this  inside the impl for `BalancerContext` 

```
impl BalancerContext {

// Other functions 

pub fn select_monolith(&self) -> anyhow::Result<&BalancerMonolith> {

    // Processing and filtering here.

    let selected = self.monolith_selection.select_monolith(Vec<&BalancerMonolith>);
    selected
    }
}
```

Could you take a look?

dyc3:
start with creating the function

```rust
fn filter_monoliths(&self) -> Vec<&BalancerMonolith> {

}
```

then, take the logic from `BalancerContext::select_monolith`, and instead of consuming those iterators with `min_by`, turn them into `Vec`s

Victor-M-Giraldo:
What exactly is the preprocessing trying to do? Is it trying to select a monolith?

The change I made does turn the iterators into vectors, but I still can't call `select_monolith()`

Also, since the preprocessing for what were originally `select_monolith` and random_monolith` are different don't we need two filter functions anyways?

dyc3:
> since the preprocessing for what were originally `select_monolith` and `random_monolith` are different don't we need two filter functions anyways?

No, they are exactly the same.


The preprocessing determines the valid candidates that should be considered when prioritizing, which is the responsibility of `trait MonolithSelection`.

Victor-M-Giraldo:
I think I'm going to keep trying tomorrow. I'm getting pretty stuck up on this. 

What I'm struggling with is that this preprocessing for `select_monolith`

```
let in_region = self
            .monoliths_by_region
            .get(BalancerConfig::get().region.as_str());
        if let Some(in_region) = in_region {
            let selected = in_region
                .iter()
                .flat_map(|id| self.monoliths.get(id))
                .min_by(|x, y| cmp(x, y));
            if let Some(s) = selected {
                return Ok(s);
            }
        }
```

is different than the one you do for `random_monolith`

```
let in_region = self
            .monoliths_by_region
            .get(BalancerConfig::get().region.as_str());
        if let Some(in_region) = in_region {
            let selected = in_region.iter().choose(&mut rand::thread_rng());
            if let Some(s) = selected {
                if let Some(m) = self.monoliths.get(s) {
                    return Ok(m);
                }
            }
        }
```

unless the preprocessing refers only to this as its the only part that is 100% the same between the two, and anything onward should be in the respective functions.
```
let in_region = self
            .monoliths_by_region
            .get(BalancerConfig::get().region.as_str());
```

dyc3:
```rust
fn filter_monoliths(&self) -> Vec<&BalancerMonolith> {
        let in_region = self
            .monoliths_by_region
            .get(BalancerConfig::get().region.as_str());
        if let Some(in_region) = in_region {
            return in_region
                .iter()
                .flat_map(|id| self.monoliths.get(id))
                .collect();
        }

        self.monoliths.values().collect()
    }
```

Victor-M-Giraldo:
Still want to write the unit tests. I will do them, but I want to get started on the other issue first.

crates/ott-balancer/src/balancer.rs
```diff
@@ -356,50 +369,19 @@ impl BalancerContext {
...
+                .collect();
         }
 
-        let selected = self
-            .monoliths
-            .values()
-            .choose(&mut rand::thread_rng())
-            .ok_or(anyhow::anyhow!("no monoliths available"))?;
-        Ok(selected)
+        self.monoliths.values().collect()
```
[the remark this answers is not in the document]
dyc3: this is not correct

crates/ott-balancer/src/service.rs
```diff
@@ -163,7 +163,10 @@ impl Service<Request<IncomingBody>> for BalancerService {
                                 );
                                 ctx_read.monoliths.get(&locator.monolith_id())
                             } else {
-                                ctx_read.select_monolith().ok()
+                                let filtered = ctx_read.filter_monoliths();
+                                let selected =
+                                    ctx_read.monolith_selection.select_monolith(filtered).ok();
+                                selected
```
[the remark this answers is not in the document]
dyc3: @Victor-M-Giraldo this suggestion is applicable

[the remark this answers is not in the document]
Victor-M-Giraldo: @coderabbitai This gives me an error with ? not being able to convert to hyper::error. Do you have another solution?

Victor-M-Giraldo:
```
ctx.add_monolith(monolith);
ctx.add_room(room_name.clone(), RoomLocator::new(monolith_id, 0))
            .expect("failed to add room");
```

Is this how I add a room to a monolith?

crates/ott-balancer/src/selection.rs
```diff
@@ -49,34 +49,72 @@ mod test {
...
+        ctx.add_room(room_one.clone(), RoomLocator::new(monolith_id_one, 0))
+            .expect("failed to add room");
+        ctx.add_room(room_two.clone(), RoomLocator::new(monolith_id_one, 0))
+            .expect("failed to add room");
+
+        ctx.add_monolith(monolith_two);
+        ctx.add_room(room_three.clone(), RoomLocator::new(monolith_id_two, 0))
+            .expect("failed to add room");
+
+        let filtered = ctx.filter_monoliths();
```
dyc3, line 112: You shouldn't be calling this in this test. The goal here is to test `MinRoomsSelector`. Any impl of `MonolithSelection`  only operates on a set of `BalancerMonolith`s.

Victor-M-Giraldo:
Like this instead?

If this is right, do you think there's a way to cut down on how many lines it is?

dyc3 (approved):

PR #1244, opened 2024-01-21
initalize visualization grafana plugin

cjrkoa:


balancer-visualization-prototype/ott-visualizationprototype-panel/package.json
```diff
@@ -0,0 +1,71 @@
+{
+  "name": "visualization-prototype",
```
dyc3, line 2: nit: Why'd you call it prototype? This is the permanent name of the package. Please make it something short, like `ott-vis`.

dyc3:
> when I try to run it just launches OTT

That's because this is a monorepo. The commands will work if you cd into the directory with the `package.json`. It needs to be added to the workspace packages. Read https://classic.yarnpkg.com/lang/en/docs/workspaces/

balancer-visualization-prototype/ott-visualizationprototype-panel/LICENSE
dyc3, line 1: Remove this

dyc3 (changes requested):
Please move the package to `packages/`, like I asked in the previous PR. Other than that, this is fine.

crates/ott-balancer-visualization/ott-visualizationprototype-panel/package.json
```diff
@@ -0,0 +1,71 @@
+{
+  "name": "ott-visualization",
```
dyc3, line 2: Please name this `ott-vis`. I do not want to type out this big long name every time I need to.

dyc3 (changes requested):
I have no idea why you moved it to `crates`, that's for rust crates, which this is not. Please move it to `packages/` and fix the folder structure so that the name of the package's folder matches the name of the package, and so that the package is not double nested. The full path to the `package.json` should be `packages/ott-vis/package.json`.

Also, you still need to add the package to the workspace in the root `package.json`.

cjrkoa:
> I have no idea why you moved it to `crates`, that's for rust crates, which this is not. Please move it to `packages/` and fix the folder structure so that the name of the package's folder matches the name of the package, and so that the package is not double nested. The full path to the `package.json` should be `packages/ott-vis/package.json`.
> 
> Also, you still need to add the package to the workspace in the root `package.json`.

I didn't see a packages folder and figured you meant crates. My bad

package.json
```diff
@@ -10,7 +10,8 @@
   "workspaces": [
     "common",
     "server",
-    "client"
+    "client",
+    "ott-vis"
```
dyc3, line 14: it needs to be the full path from the root of the repo
```suggestion
    "packages/ott-vis"
```

dyc3 (changes requested):

dyc3 (approved):

dyc3:
@coderabbitai Open an issue for updating the CI pipelines to add this package, and the release pipelines to sign the plugin

PR #1253, opened 2024-01-22
add separate toggles for auto-skip categories

### LINKED ISSUE #1209: separate toggles for auto-skip categories

I want to auto-skip sponsors but not intros. 
The `Auto-skip sponsored segments, intros, self-promos, etc. using SponsorBlock data.` checkbox does not allow for this behavior.

I would like one checkbox for each type of sponsor block data and one parent checkbox to disable/enable all of them at once for people who like the current behavior.

From the `videoSlider` colors it looks like the type of each auto-skip segment is already known.

If there is no work on this already I would like to implement this myself.

### Additional context

_No response_

ceofil:
Resolves #1209


server/models/room.ts
```diff
@@ -18,6 +18,7 @@ export interface RoomAttributes {
 	"role-mod": Array<number>;
 	"role-trusted": Array<number>;
 	"autoSkipSegments": boolean;
+	"autoSkipSegmentCategories": SegmentCategories;
```
dyc3, line 21: I would prefer if this replaced the boolean column instead of getting added alongside it. It'll make the db migration a little bit more complex, but I think it'll be worth it for the reduced complexity in the logic.

common/models/types.ts
```diff
@@ -53,6 +54,10 @@ export type ClientInfo = {
 	status?: PlayerStatus;
 };
 
+export type SegmentCategories = {
+	[K in Category]: boolean;
+}
```
dyc3, line 59: It would be slightly more efficient disk space wise if this type was an array of `Category` instead of an object. Maybe not so much in postgres (because of jsonb), but definitely in redis. On the other hand, this approach doesn't require us to use `Set` to guarantee uniqueness of items, which would add to the complexity.

What's your opinion?

docs/api.yaml
```diff
@@ -106,6 +106,21 @@ paths:
...
+                        outro: 
+                          type: boolean
+                        interaction: 
+                          type: boolean
+                        selfpromo: 
+                          type: boolean
+                        music_offtopic: 
+                          type: boolean
+                        preview: 
+                          type: boolean
```
dyc3, line 123: Thanks for updating the docs, good attention to detail. Make this a reusable component to reduce the duplicated bits.

client/src/locales/en.ts
```diff
@@ -244,7 +244,14 @@ export default {
 		"dj-hint":
 			"When the video ends, start the same video from the beginning. Good for looping background music.",
 		"auto-skip-text":
-			"Auto-skip sponsored segments, intros, self-promos, etc. using SponsorBlock data.",
+			"Auto-skip SponsorBlock data.",
```
dyc3, line 247: nit
```suggestion
			"Auto-skip undesirable video segments using SponsorBlock data.",
```

dyc3 (changes requested):
Thanks for the high quality PR! Love to see it.

The only major blocker I see is that there's no validation on the `SegmentCategories` before it is applied to the Room and saved in redis or database. A malicious client could DoS because they could add a bunch of items to the object and eat up a bunch of memory, and/or disk space. Preferably, this should be done in the handler for `RoomRequest.ApplySettingsRequest`.

Also, I'd like to have unit tests for storing and retrieving this new column in the database.

common/models/types.ts
ceofil: Totally agree. I completely forgot about the redis layer.

client/src/components/RoomSettingsForm.vue
```diff
@@ -65,55 +65,54 @@
...
+						:disabled="!granted('configure-room.other')"
+						data-cy="input-auto-skip-music_offtopic"
+					/>
+					<v-checkbox
+						v-model="autoSkipSegmentCategoriesObject.preview"
+						:label="$t('room-settings.auto-skip-text-preview')"
+						:disabled="!granted('configure-room.other')"
+						data-cy="input-auto-skip-preview"
+					/>
+				</div>
```
dyc3, line 115: I think something like this might work better for the UI

image
See: https://vuetifyjs.com/en/components/selects/#chips
Or [this playground](https://play.vuetifyjs.com/#eNqNT9EKgzAM/JWQFzdwytjGQFQY+wz1oWpkhTqLrTIU/31tVdjjni53ySWXbEbVV+FDymAcCCOMNbVSME1p/gaIx5MiQZW2BGA8tV1NIslxZGKgHFc54sakjOpwV6sXl2otBSud6+mkrd8OQnMpyLI0DvdL5m4c/oQwVFU9l7YBQB/Z9Rpqapixw7yuqplmERyOkKRw2DQAlyaCzGu6zvPBK1lvoeHT5OhgsPD3affRn9PL0RWLjbqFw8XHS3AN7mjxFpyx+AIyxGr8)

common/models/rest-api.ts
```diff
@@ -53,10 +54,9 @@ export interface OttApiResponseGetRoom extends RoomSettings {
 	queue: QueueItem[];
 	hasOwner: boolean;
 	grants: Grants;
-	autoSkipSegmentCategories: SegmentCategories;
+	autoSkipSegmentCategories: Array<Category>;
```
dyc3, line 57: nit
```suggestion
	autoSkipSegmentCategories: Category[];
```

server/room.ts
```diff
@@ -227,16 +224,9 @@ export class Room implements RoomState {
...
-	_autoSkipSegmentCategories: SegmentCategories = {
-        'sponsor': true,
-        'intro': true,
-        'outro': true,
-        'interaction': true,
-        'selfpromo': true,
-        'music_offtopic': true,
-        'preview': true
-	};
+	_autoSkipSegmentCategories: Array<Category> = Array.from([
```
dyc3, line 227: nit
```suggestion
	_autoSkipSegmentCategories: Category[] = Array.from([
```

server/room.ts
```diff
@@ -403,20 +392,11 @@ export class Room implements RoomState {
...
-		return this._autoSkipSegments;
-	}
-	
-	public set autoSkipSegments(value: boolean) {
-		this._autoSkipSegments = value;
-		this.markDirty("autoSkipSegments");
-	}
-
-	public get autoSkipSegmentCategories(): SegmentCategories {
+	public get autoSkipSegmentCategories(): Array<Category> {
```
dyc3, line 395: nit
```suggestion
	public get autoSkipSegmentCategories(): Category[] {
```

server/room.ts
```diff
@@ -403,20 +392,11 @@ export class Room implements RoomState {
...
 		return this._autoSkipSegmentCategories;
 	}
 	
-	public set autoSkipSegmentCategories(value: SegmentCategories) {
+	public set autoSkipSegmentCategories(value: Array<Category>) {
```
dyc3, line 399: nit
```suggestion
	public set autoSkipSegmentCategories(value: Category[]) {
```

dyc3 (changes requested):
Looking good so far.

This still needs to validate incoming settings changes for `autoSkipSegmentCategories`, like I explained in my previous review.

server/room.ts
```diff
@@ -1559,6 +1560,13 @@ export class Room implements RoomState {
 				}
 			}
 		}
+		
+		if (request.settings.autoSkipSegmentCategories){
+			const autoSkipSegmentCategoriesSet = new Set(request.settings.autoSkipSegmentCategories)
+			request.settings.autoSkipSegmentCategories = ALL_SKIP_CATEGORIES.filter(
+				category => autoSkipSegmentCategoriesSet.has(category)
+			)
+		}
```
dyc3, line 1569: It would be nice to have a couple unit tests for this behavior.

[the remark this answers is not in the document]
ceofil: Currently we have to do `len(ALL_SKIP_CATEGORIES)` lookups in `autoSkipSegmentCategoriesSet`. 
I'm not sure how `_.isEqual` works but with the suggested approach we have `len(autoSkipSegmentCategoriesSet)` lookups in `autoSkipSegmentCategoriesSet`

- `len(ALL_SKIP_CATEGORIES)` is 7
- `len(autoSkipSegmentCategoriesSet)` can be anything

server/tests/unit/api/room.spec.ts
```diff
@@ -370,4 +370,79 @@ describe("Room API", () => {
...
+				email: userEmail,
+				username: "owner2",
+				password: "password1234",
+			});
+			getSessionInfoSpy = jest.spyOn(tokens, "getSessionInfo").mockResolvedValue({
+				isLoggedIn: true,
+				user_id: user.id,
+			});
+			validateSpy = jest.spyOn(tokens, "validate").mockResolvedValue(true);
+			conf.set("room.enable_create_permanent", true);
```
dyc3, line 392: The default for this option is true, so setting it is unnecessary here.

dyc3 (changes requested):
I'll be able to give you better feedback once all the lints are fixed.

ceofil:
When I run `yarn lint` locally i get a bunch of errors regarding "describe", "it" and "expect" not being defined and then it exits with code 1. 

Is there something I should change in the config? 

<details><summary>Output</summary>
<p>
....

/home/teofil/projects/opentogethertube/common/ts-out/tests/unit/voteskip.spec.js
   2:1  error  'describe' is not defined  no-undef
   3:2  error  'it' is not defined        no-undef
  21:3  error  'expect' is not defined    no-undef

✖ 110 problems (110 errors, 0 warnings)

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed.
Exit code: 1
Command: /usr/bin/node
Arguments: /usr/local/lib/node_modules/yarn/lib/cli.js run lint
Directory: /home/teofil/projects/opentogethertube/common
Output:

info Visit https://yarnpkg.com/en/docs/cli/workspaces for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

</p>
</details>

dyc3:
Ah yeah I've fixed that on `master`. You can either rebase/merge `master` or delete the directories `common/ts-out` and `server/ts-out`

server/room.ts
```diff
@@ -224,7 +225,7 @@ export class Room implements RoomState {
 	_owner: User | null = null;
 	grants: Grants = new Grants();
 	userRoles: Map<Role, Set<number>>;
-	_autoSkipSegments = true;
+	_autoSkipSegmentCategories: Array<Category> = Array.from([]);
```
dyc3, line 228: This makes the default value for newly created rooms (both temporary and permanent) empty, when it should be all the categories.

dyc3 (changes requested):
When I was doing manual testing using the dev server, I couldn't get in a room because of this error.

image

However, building the client manually it worked fine. I suspect this is yarn being stupid again. Does it work for you when using `yarn run dev`?

ceofil:
> When I was doing manual testing using the dev server, I couldn't get in a room because of this error.
> 
> image
> 
> However, building the client manually it worked fine. I suspect this is yarn being stupid again. Does it work for you when using `yarn run dev`?

Yup, everything works fine when I run `yarn run dev`

dyc3:
I've resolved the import issues with `ott-common`. I don't have perms to push to this branch, can you merge master into here again?

dyc3 (approved):
Looks good! Just merge master into here and I'll be able to get this deployed safely.

PR #1272, opened 2024-01-26
client: show playback errors in the UI

dyc3:
Currently, this doesn't show any error when the server responds with a failure. Strangely, the video element does not raise the `error` event when this occurs, nor does it set the `error` field on the element.

### DIFF
--- a/client/src/components/players/OmniPlayer.vue
+++ b/client/src/components/players/OmniPlayer.vue
@@ -19,6 +19,23 @@
 					</div>
 				</v-container>
 			</v-sheet>
+			<v-sheet color="error" v-if="playbackError !== undefined">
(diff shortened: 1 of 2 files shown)


client/src/components/players/OmniPlayer.vue
```diff
@@ -19,6 +19,23 @@
 					</div>
 				</v-container>
 			</v-sheet>
+			<v-sheet color="error" v-if="playbackError !== undefined">
+				<v-container fluid style="padding: 6px">
+					<div style="display: flex; align-items: center">
+						<v-icon>mdi:alert-circle</v-icon>
```
dyc3, line 25: this should be changed to use font awesome

dyc3:
This PR is too stale now.

PR #1287, opened 2024-01-30
make monolith ids constant

### LINKED ISSUE #1251: make monolith ids constant from the perspective of the balancers
Currently, monoliths are assigned a unique `MonolithId` by balancers. However, this makes it extremely difficult to correlate which monolith is which just by querying the balancers. 

- Change the monolith so that it generates it's own id, and conveys it to the balancer in the init message.
- Change the balancer to use this id instead of generating it.
- Add unit tests to the monolith to assert for this new behavior.
    - useful related docs: https://jestjs.io/docs/mock-functions
    - see tests in `clientmanager.spec.ts`

Also update the spec doc to reflect these changes.

Victor-M-Giraldo:
closes #1251


dyc3:
Yeah that sounds right

Victor-M-Giraldo:
How is the id defined in the constructor for `BalancerConnection` any different than a `MonolithID`?

Also, is there an example somewhere in the code of reading the payload, or how would I access that id to use in `connection.rs` rather than generating it in there?

server/balancer.ts
```diff
@@ -76,6 +76,7 @@ class BalancerManager {
 			payload: {
 				port: conf.get("port"),
 				region: conf.get("balancing.region"),
+				id: uuidv4(),
```
dyc3, line 79: This still makes the id random for every balancer it connects to. It needs to be constant across reconnects. It's ok if the id is not constant across service restarts (eg don't store it in a file, it being a global is fine).

crates/ott-balancer/src/connection.rs
```diff
@@ -126,18 +123,21 @@ async fn connect_and_maintain(
 
         // Handle connection initialization
         let mut outbound_rx;
+        let id_clone;
```
dyc3, line 126: nit: 
```suggestion
        let monolith_id;
```

Victor-M-Giraldo:
Having some trouble writing the test for this. Might ask for help tomorrow.

dyc3:
The most valuable test here would probably be a unit test in `clientmanager.spec.ts` that creates 2 `BalancerConnectionMock` and see if they both get sent the same id.

Victor-M-Giraldo:
How can I run only tests for a certain file? Is that possible?

dyc3:
Yeah should be possible. Look up the docs for `jest` or `cargo test`, depending if you're working with ts or rust, respectively. I can never get it to work reliably with jest

Victor-M-Giraldo:
Sounds good. I'll take a look. Just want to make sure the test is working without running the entire suite.

Also, the monolith IDS should be different, correct?

dyc3:
Monolith ids should be the same within the context of the monolith. Across multiple monoliths, their ids should be different.

Victor-M-Giraldo:
Okay, that's what I thought.

Victor-M-Giraldo:
Figure this probably isn't what you mean. 

Is it something to do with `sendMock`?

server/tests/unit/clientmanager.spec.ts
```diff
@@ -172,3 +172,15 @@ describe("BalancerManager", () => {
 		expect(balancerManager.balancerConnections).not.toContain(con2);
 	});
 });
+
+describe("MonolithId", () => {
+	it("should have unique IDS for each instance of a monolith", () => {
```
dyc3, line 177: From this context, there is only 1 monolith: the one that is running. The client manager should send the same id to both of these mock connections.

Victor-M-Giraldo:
I only see stuff about the client ID there. Or is that what you meant?

dyc3:
No, I'm not talking about the client id, I'm talking about the goal of this pr.

Victor-M-Giraldo:
Alright.

Victor-M-Giraldo:
Think that works.

server/tests/unit/clientmanager.spec.ts
```diff
@@ -181,10 +181,11 @@ describe("MonolithId", () => {
 		balancerManager.addBalancerConnection(mock1);
 		balancerManager.addBalancerConnection(mock2);
 
-		const id1 = mock1.sendMock.mock.calls[0][0].payload.id;
-		const id2 = mock2.sendMock.mock.calls[0][0].payload.id;
+		const id1 = mock1.sendMock.mock.calls[0][0].filter((msg: MsgM2B) => msg.type == "init")[0]
+			.payload.id;
+		const id2 = mock2.sendMock.mock.calls[0][0].filter((msg: MsgM2B) => msg.type == "init")[0]
+			.payload.id;
```
dyc3, line 187: this is filtering the wrong one. you need to filter the calls, not the call arguments

dyc3 (changes requested):

server/tests/unit/clientmanager.spec.ts
dyc3: It's ok to leave it the way it was

dyc3:
if you want to rebase on master, I just pushed a commit that adds type annotations to `sendMock`

server/tests/unit/clientmanager.spec.ts
```diff
@@ -201,3 +202,18 @@ describe("BalancerManager", () => {
...
+		const mock2 = new BalancerConnectionMock();
+
+		balancerManager.addBalancerConnection(mock1);
+		balancerManager.addBalancerConnection(mock2);
+
+		const id1 = (mock1.sendMock.mock.calls[0][0].payload as M2BInit).id;
+		const id2 = (mock2.sendMock.mock.calls[0][0].payload as M2BInit).id;
+
+		expect(id1).toEqual(id2);
+	});
```
dyc3, line 218: this should also assert that id1 and id2 are both not undefined

dyc3 (changes requested):

dyc3 (approved):

PR #1292, opened 2024-02-03
add panels to dash

### LINKED ISSUE #1280: vis: add other panels to provisioned dashboard
Add the other panels that can use grafana's built in graphs to the auto-provisioned grafana dashboard at `packages/ott-vis/provisioning/dashboards/dashboard.json`

The specifics of these panels is in the spec doc.

Victor-M-Giraldo:
closes #1280


Victor-M-Giraldo:
How do we get these panels to actually display meaningful data and not just a static graph?

dyc3:
It just requires changing the data source. Should be a trivial change for the most part. I can show the group when we meet next in person.

packages/ott-vis/provisioning/dashboards/dashboard.json
```diff
@@ -18,38 +18,273 @@
...
+					"placement": "bottom",
+					"showLegend": true
+				},
+				"tooltip": {
+					"mode": "single",
+					"sort": "none"
 				}
-			],
-			"title": "Panel Title",
-			"type": "ott-vis-panel"
```
dyc3, line 52: We need to keep this panel, ideally in the same spot it was before.

dyc3 (changes requested):
This removes our custom panel, which we need for our development workflow.

packages/ott-vis/provisioning/dashboards/dashboard.json
Victor-M-Giraldo: Added it back. What is this one going to represent though? Didn't see it line up with anything in the doc which is why I deleted it.

dyc3 (changes requested):
Please lay them out better so you don't need to scroll to see the custom panel. The custom panel should be at the top.

Victor-M-Giraldo:
Moved them around. What's the custom panel going to represent? Is it the default/region views? I deleted it originally because I didn't see what it lined up with in the doc.

dyc3 (approved):

dyc3:
Yes, its going to be the default/region views.

PR #1303, opened 2024-02-08
add a command line argument to validate the config and exit

### LINKED ISSUE #1297: monolith: add a command line argument to simply validate the config and exit

I want to be able to validate configs before they reach production and start causing the service to crash.

Add a `--validate` cli arg to the monolith that will load the config specified and if it succeeds without errors, exit immediately with code 0. Otherwise, exit with a non-zero exit code.

### Additional context

_No response_

Victor-M-Giraldo:
closes #1297

### DIFF
--- a/server/app.ts
+++ b/server/app.ts
@@ -54,7 +54,7 @@ export async function main() {
 	await buildClients();
 	buildRateLimiter();
 
-	if (process.argv.includes("--validate")) {
+	if (process.argv.includes("--validate") || process.argv.includes("-v")) {
 		validateConfig();
 	}
 
--- a/server/app.ts
+++ b/server/app.ts
@@ -54,6 +54,16 @@ export async function main() {
 	await buildClients();
 	buildRateLimiter();
 
+	if (process.argv.includes("--validate")) {
+		try {
+			loadConfigFile();
+		} catch (e) {
+			log.error(e);
+			process.exit(1);
+		}
+		process.exit(0);
+	}
+
 	if (
 		conf.get("env") === "production" &&
 		conf.get("db.mode") !== "sqlite" &&
--- a/server/app.ts
+++ b/server/app.ts
@@ -34,9 +34,11 @@ export async function main() {
 	if (process.argv.includes("--validate")) {
 		let result = validateConfig();
 		if (!result.ok) {
+			log.error("Config validation failed:");
 			log.error(result.value.message);
 			process.exit(1);
 		} else {
+			log.info("Config validation passed");
 			process.exit(0);
 		}
 	}
--- a/server/app.ts
+++ b/server/app.ts
@@ -22,7 +22,7 @@ import { initExtractor } from "./infoextractor";
 import session, { SessionOptions } from "express-session";
 import RedisStore from "connect-redis";
 import { setupPostgresMetricsCollection } from "./storage.metrics";
-import { stderr } from "process";
+
 
 const app = express();
 
--- a/server/app.ts
+++ b/server/app.ts
@@ -32,7 +32,9 @@ export async function main() {
 	loadConfigFile();
 	setLogLevel(conf.get("log.level"));
 	if (process.argv.includes("--validate")) {
-		if (!validateConfig()) {
+		let result = validateConfig();
+		if (!result.ok) {
+			log.error(result.value.message);
 			process.exit(1);
 		} else {
 			process.exit(0);
--- a/server/app.ts
+++ b/server/app.ts
@@ -31,6 +31,9 @@ export async function main() {
 	setLogger(getLogger("config"));
 	loadConfigFile();
 	setLogLevel(conf.get("log.level"));
+	if (process.argv.includes("--validate")) {
+		validateConfig();
+	}
 
 	const env = conf.get("env");
 	const heroku = conf.get("heroku");
@@ -53,10 +56,6 @@ export async function main() {
 	loadModels();
 	await buildClients();
 	buildRateLimiter();
-
-	if (process.argv.includes("--validate")) {
-		validateConfig();
-	}
 
 	if (
 		conf.get("env") === "production" &&
--- a/server/app.ts
+++ b/server/app.ts
@@ -16,12 +16,13 @@ import websockets from "./websockets";
 import clientmanager from "./clientmanager";
 import roommanager from "./roommanager";
 import bodyParser from "body-parser";
-import { loadConfigFile, conf, setLogger } from "./ott-config";
+import { loadConfigFile, conf, setLogger, validateConfig } from "./ott-config";
 import { buildRateLimiter } from "./rate-limit";
 import { initExtractor } from "./infoextractor";
 import session, { SessionOptions } from "express-session";
 import RedisStore from "connect-redis";
 import { setupPostgresMetricsCollection } from "./storage.metrics";
+import { stderr } from "process";
 
 const app = express();
 
@@ -55,13 +56,7 @@ export async function main() {
 	buildRateLimiter();
 
 	if (process.argv.includes("--validate")) {
-		try {
-			conf.validate({ allowed: "strict" });
-		} catch (e) {
-			log.error(e);
-			process.exit(1);
-		}
-		process.exit(0);
+		validateConfig();
 	}
 
 	if (
--- a/server/app.ts
+++ b/server/app.ts
@@ -32,7 +32,11 @@ export async function main() {
 	loadConfigFile();
 	setLogLevel(conf.get("log.level"));
 	if (process.argv.includes("--validate")) {
-		validateConfig();
+		if (!validateConfig()) {
+			process.exit(1);
+		} else {
+			process.exit(0);
+		}
 	}
 
 	const env = conf.get("env");
--- a/server/app.ts
+++ b/server/app.ts
@@ -54,7 +54,7 @@ export async function main() {
 	await buildClients();
 	buildRateLimiter();
 
-	if (process.argv.includes("--validate") || process.argv.includes("-v")) {
+	if (process.argv.includes("--validate")) {
 		validateConfig();
 	}
 
--- a/server/app.ts
+++ b/server/app.ts
@@ -23,7 +23,6 @@ import session, { SessionOptions } from "express-session";
 import RedisStore from "connect-redis";
 import { setupPostgresMetricsCollection } from "./storage.metrics";
 
-
 const app = express();
 
 export async function main() {
--- a/server/app.ts
+++ b/server/app.ts
@@ -55,8 +55,9 @@ export async function main() {
 	buildRateLimiter();
 
 	if (process.argv.includes("--validate")) {
+		loadConfigFile();
 		try {
-			loadConfigFile();
+			conf.validate({ allowed: "strict" });
 		} catch (e) {
 			log.error(e);
 			process.exit(1);
--- a/server/app.ts
+++ b/server/app.ts
@@ -55,7 +55,6 @@ export async function main() {
 	buildRateLimiter();
 
 	if (process.argv.includes("--validate")) {
-		loadConfigFile();
 		try {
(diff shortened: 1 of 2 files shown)


dyc3:
Go ahead and fix your PR title

server/app.ts
```diff
@@ -54,6 +54,16 @@
...
 
+	if (process.argv.includes("--validate")) {
+		try {
+			loadConfigFile();
+		} catch (e) {
+			log.error(e);
+			process.exit(1);
+		}
+		process.exit(0);
+	}
```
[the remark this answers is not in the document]
dyc3: Yeah @Victor-M-Giraldo this should happen before the config gets loaded at the top

server/app.ts
```diff
@@ -54,6 +54,16 @@ export async function main() {
 	await buildClients();
 	buildRateLimiter();
 
+	if (process.argv.includes("--validate")) {
+		try {
+			loadConfigFile();
```
dyc3, line 59: `loadConfigFile` doesn't throw exceptions when validation fails. This always succeeds and the process will exit with code 0.

dyc3 (changes requested):

server/app.ts
dyc3: Just to clarify -- include a key in your config that doesn't exist.
```toml
foo = "bar"
```
When you run `yarn start --validate` the process will exit with code 0.

Victor-M-Giraldo: Yea, I did that. I thought it wasn't working. I just wanted confirmation.

So what I should do is check if the schema for `.env` files doesn't match up with the `content` in `loadConfigFile` and throw a `log.error` if it does?

dyc3: You can verify it yourself. `echo $?` will show you the exit code of the previous command that was run.

See if `conf.validate` does what we want. It would be nice to also take into account the logic in `validateMail` too.

Victor-M-Giraldo: I believe if we change this line `conf.validate({ allowed: "warn" });` to `conf.validate({ allowed: "strict" });` in `loadConfigFile` it will work. Checking now.

server/app.ts
```diff
@@ -54,6 +54,17 @@ export async function main() {
 	await buildClients();
 	buildRateLimiter();
 
+	if (process.argv.includes("--validate")) {
+		loadConfigFile();
```
dyc3, line 58: This loads the config file twice.

dyc3 (changes requested):

Victor-M-Giraldo:
For clarification, what changes do you want with `validateMail`? 

As is `--validate` works right now the way you described. Any keys defined that are not in the schema throw an error.

dyc3:
Think about the use case. I want to avoid deploying configs that break functionality. 

Email is critical for letting users reset their passwords. Failing to have that properly configured so that it works (when it is desired by the deployer) is bad.

server/ott-config.ts
```diff
@@ -487,6 +487,17 @@ export function loadConfigFile() {
 	conf.validate({ allowed: "warn" });
 }
 
+export function validateConfig() {
+	try {
+		conf.validate({ allowed: "strict" });
+		validateMail();
```
dyc3, line 493: `validateMail` intentionally doesn't throw exceptions to not block startup. This is a feature. However, that also means that any validation errors will not be caught by this `try`.

I would recommend making `validateMail()` return a `Result` from `ott-common`, and return `err(msg)` whenever there is an error instead of logging it.

dyc3 (changes requested):

Victor-M-Giraldo:
When using `--validate` it always exits with code 0 with these changes. I can't find out why. I feel like it's something simple I'm missing.

dyc3:
Make sure you are checking the `Result`. It wont throw an exception.

dyc3 (changes requested):
Looking good, but could be better.

server/ott-config.ts
```diff
@@ -489,14 +489,20 @@ export function loadConfigFile() {
...
 		process.exit(1);
+	} else {
+		try {
+			conf.validate({ allowed: "strict" });
+		} catch (e) {
+			log.error(e);
+			process.exit(1);
+		}
+
+		process.exit(0);
```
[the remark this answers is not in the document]
Victor-M-Giraldo: Yeah, I took that into consideration when doing this. I just couldn't figure out the best way to do it. Will keep trying.

server/ott-config.ts
```diff
@@ -536,10 +542,9 @@ function postProcessConfig(): void {
 	}
 
 	if (conf.get("mail.enabled")) {
-		try {
-			validateMail();
-		} catch (e) {
-			log.error(e);
+		const result = validateMail();
+		if (!result.ok) {
+			log.error(result.value.message);
```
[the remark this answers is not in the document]
Victor-M-Giraldo: That's a pretty easy fix, will do that now.

Victor-M-Giraldo:
I think this is way better. I tested it and it looks good. The only problem I'm seeing is with repeat error messages for the mail config. It's because of this in `postProcessConfig()`. Is it alright if I get rid of that?

```
if (conf.get("mail.enabled")) {
		const result = validateMail();
		if (!result.ok) {
			conf.set("mail.enabled", false);
			log.error(result.value.message);
		}
	}
```

server/app.ts
```diff
@@ -54,7 +54,7 @@ export async function main() {
 	await buildClients();
 	buildRateLimiter();
 
-	if (process.argv.includes("--validate")) {
+	if (process.argv.includes("--validate") || process.argv.includes("-v")) {
```
dyc3, line 57: Lets leave the short alias out for now. `-v` is usually short for `--version`, and I want to minimize breaking changes in the future.

```suggestion
	if (process.argv.includes("--validate")) {
```

server/ott-config.ts
```diff
@@ -489,20 +489,24 @@ export function loadConfigFile() {
...
-		} catch (e) {
-			log.error(e);
-			process.exit(1);
+	const mailEnabled = conf.get("mail.enabled");
+	const mailResult = validateMail();
+	const confResult = validateConf();
+
+	if ((!mailResult.ok && mailEnabled) || !confResult.ok) {
+		if (!mailResult.ok) {
+			conf.set("mail.enabled", false);
```
dyc3, line 498: No need for this. The process is exiting anyway.
```suggestion
```

server/ott-config.ts
```diff
@@ -554,23 +559,29 @@ function postProcessConfig(): void {
...
 }
 
+function validateConf(): Result<void, Error> {
+	try {
+		conf.validate({ allowed: "strict" });
+		return ok(undefined);
+	} catch (e) {
+		return err(e);
+	}
+}
```
dyc3, line 569: Use `intoResult()` for this.

dyc3 (changes requested):

server/ott-config.ts
Victor-M-Giraldo: Trying to do so now but having some trouble. I see how the two are super similar where `conf.validate()` lines up with the result in `intoResult()`

I'm trying to do `intoResult(conf.validate({allowed: "strict"})`, but it isn't working since I think the `func()` in `intoResult()` takes no args.

Hint or maybe some clarification on if I'm right?

dyc3: `intoResult` takes a function. `intoResult(() => foo())`

Victor-M-Giraldo: Why doesn't something like this work?

```
function validateConf() {
	conf.validate({ allowed: "strict" });
}

const confResult = intoResult(validateConf());
```

Victor-M-Giraldo: Nevermind, it does. Dumb mistake.

server/ott-config.ts
```diff
@@ -524,7 +545,11 @@ function postProcessConfig(): void {
 	}
 
 	if (conf.get("mail.enabled")) {
-		validateMail();
+		const result = validateMail();
+		if (!result.ok) {
+			conf.set("mail.enabled", false);
+			log.error(result.value.message);
```
dyc3, line 551:
```suggestion
			log.error(result.value.message);
			log.error("Disabling mail");
			conf.set("mail.enabled", false);
```

server/ott-config.ts
```diff
@@ -533,24 +558,24 @@ function postProcessConfig(): void {
...
 
-function validateMail() {
+function validateMail(): Result<void, Error> {
 	const sender = conf.get("mail.sender_email");
+
 	if (sender && !validator.isEmail(sender)) {
-		log.error(`Invalid email address ${sender} found in config, disabling mail.`);
-		conf.set("mail.enabled", false);
-		return;
+		return err(new Error(`Invalid email address ${sender} found in config, disabling mail.`));
```
dyc3, line 565:
```suggestion
		return err(new Error(`Invalid email address ${sender} found in config.`));
```

server/ott-config.ts
```diff
@@ -533,24 +558,24 @@ function postProcessConfig(): void {
...
 	}
 
 	if (!conf.get("mail.mailjet_api_key") || !conf.get("mail.mailjet_api_secret")) {
-		log.error("Mailjet API key and secret are required to send mail, disabling mail.");
-		conf.set("mail.enabled", false);
-		return;
+		return err(
+			new Error("Mailjet API key and secret are required to send mail, disabling mail.")
```
dyc3, line 570:
```suggestion
			new Error("Mailjet API key and secret are required to send mail.")
```

server/ott-config.ts
```diff
@@ -533,24 +558,24 @@ function postProcessConfig(): void {
...
+		);
 	}
 
 	if (conf.get("env") === "test") {
-		log.warn("Mail is enabled in test mode. Forcing sandbox mode.");
-		conf.set("mail.mailjet_sandbox", true);
+		return err(new Error("Mail is enabled in test mode. Forcing sandbox mode."));
 	}
```
dyc3, line 576: This is not equivalent functionality. Move this check somewhere else and restore the original behavior.

dyc3 (changes requested):

server/app.ts
```diff
@@ -54,6 +54,10 @@ export async function main() {
 	await buildClients();
 	buildRateLimiter();
 
+	if (process.argv.includes("--validate")) {
+		validateConfig();
+	}
```
dyc3, line 59: This needs to happen before the config gets used to do anything. For example, if you set `redis.host` to a valid host, but the host is not available, this will hang on `buildClients()` retrying forever until it connects to redis. Move it to right under the `setLogLevel` call

server/ott-config.ts
```diff
@@ -487,6 +488,26 @@ export function loadConfigFile() {
...
+	if ((!mailResult.ok && mailEnabled) || !confResult.ok) {
+		if (!mailResult.ok) {
+			log.error(mailResult.value.message);
+		}
+
+		if (!confResult.ok) {
+			log.error(confResult.value.message);
+		}
+
+		process.exit(1);
```
dyc3, line 505: On it's own, `validateConfig()` does not make it clear to the reader that calling it will make the process exit. Move the calls to `process.exit()` to `app.ts`

dyc3 (changes requested):

Victor-M-Giraldo:
I don't know if it's better to make `validateConfig()` return a bool or a `Result <T , E>` to check the return value. Do you have thoughts on that?

dyc3:
A `Result` please

server/ott-config.ts
```diff
@@ -488,7 +488,7 @@ export function loadConfigFile() {
 	conf.validate({ allowed: "warn" });
 }
 
-export function validateConfig() {
+export function validateConfig(): boolean {
```
dyc3: Go ahead and make this a `Result`

dyc3 (changes requested):

PR #1317, opened 2024-02-11
balancer: add a command line argument to validate config

### LINKED ISSUE #1296: balancer: add a command line argument to simply validate the config and exit

I want to be able to validate configs before they reach production and start causing the service to crash.

Add a `--validate` cli arg to the balancer that will load the config specified and if it succeeds without errors, exit immediately with code 0. Otherwise, exit with a non-zero exit code.

### Additional context

_No response_

Victor-M-Giraldo:
closes #1296

### DIFF
--- a/crates/ott-balancer/src/lib.rs
+++ b/crates/ott-balancer/src/lib.rs
@@ -34,7 +34,7 @@ pub async fn run() -> anyhow::Result<()> {
     let config = BalancerConfig::get();
 
     let _validate = Arg::new("validate")
-        .short('c')
+        .short('v')
         .long("validate")
         .help("Validate the configuration and exit");
 
--- a/crates/ott-balancer/src/lib.rs
+++ b/crates/ott-balancer/src/lib.rs
@@ -31,7 +31,6 @@ pub async fn run() -> anyhow::Result<()> {
     let args = config::Cli::parse();
 
     let loaded_config = BalancerConfig::load(&args.config_path);
-    let config = BalancerConfig::get();
 
     if args.validate {
         match loaded_config {
@@ -39,11 +38,13 @@ pub async fn run() -> anyhow::Result<()> {
                 std::process::exit(0);
             }
             Err(err) => {
-                error!("Error loading configuration: {:?}", err);
+                eprintln!("Error loading configuration: {:?}", err);
                 std::process::exit(1);
             }
         }
     }
+
+    let config = BalancerConfig::get();
 
     let console_layer = if args.remote_console {
         console_subscriber::ConsoleLayer::builder()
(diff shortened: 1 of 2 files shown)


crates/ott-balancer/src/lib.rs
```diff
@@ -33,6 +33,23 @@ pub async fn run() -> anyhow::Result<()> {
     BalancerConfig::load(&args.config_path)?;
     let config = BalancerConfig::get();
 
+    let _validate = Arg::new("validate")
+        .short('v')
+        .long("validate")
+        .help("Validate the configuration and exit");
```
dyc3, line 39: Why is this not in `config::Cli`? This makes it so that the help text doesn't include the flag.

crates/ott-balancer/src/lib.rs
```diff
@@ -33,6 +33,23 @@ pub async fn run() -> anyhow::Result<()> {
...
+
+    if std::env::args().any(|arg| arg == "--validate") {
+        match BalancerConfig::load(&args.config_path) {
```
dyc3, line 42: this is double loading the config

dyc3 (changes requested):

crates/ott-balancer/src/lib.rs
Victor-M-Giraldo: I figured that out after looking at the docs again. I was just about to change that.

Victor-M-Giraldo:
This isn't working yet. I think its because `load` returns an `Ok(())` so it always exits with code 0.

Victor-M-Giraldo:
Yeah, I think I'm going to need some help with this, or at least some direction.

The two ideas I had were making a function similar to `postProcessConfig` in `ott-config.ts` where if the discovery method was with fly for example it checked that the `monolith_port` and `fly_app` were defined. Don't think that's the right way to go though.

The other was messing around with `load`.

dyc3:
Have you looked at the docs for the `figment` crate yet?

Victor-M-Giraldo:
Yes, but I was having some trouble understanding them. I'll take another look though and let you know.

Victor-M-Giraldo:
Going to add messages indicating whether config validation succeeded or failed when I get home.

crates/ott-balancer/src/lib.rs
```diff
@@ -35,9 +35,11 @@ pub async fn run() -> anyhow::Result<()> {
     if args.validate {
         match loaded_config {
             Ok(()) => {
+                println!("Configuration is valid");
```
dyc3, line 38:
```suggestion
                eprintln!("Configuration is valid");
```

crates/ott-balancer/src/lib.rs
```diff
@@ -35,9 +35,11 @@ pub async fn run() -> anyhow::Result<()> {
...
                 std::process::exit(0);
             }
             Err(err) => {
+                println!("Configuration is invalid");
```
dyc3, line 42:
```suggestion
                eprintln!("Configuration is invalid");
```

dyc3 (approved):

PR #1335, opened 2024-02-14
ci: attempt to make cypress component tests more reliable

dyc3:


dyc3:
@coderabbitai open an issue to fix this flaky test: `tests/unit/storage.spec.ts > Storage: Room Spec > should return room object without extra properties`. It only seems to happen in CI and it only started to be flaky when we switched to vitest from jest.

dyc3:
After running the CI job a few times, I'm pretty confident this is will fix the issue for good.

Issue #1326, opened 2024-02-12
Get rid of Jest for good. Switch all unit tests to vitest.


dyc3:
### Is your feature request related to a problem? Please describe.

Jest and babel have eaten too many hours of my life.

### Describe the solution you'd like

We should be able to completely replace jest with vitest

PR #1347, opened 2024-02-17
switch `ott-vis` to vitest

dyc3:
related: #1326

currently half working


dyc3:
Not relevant anymore because the package name has changed

PR #1369, opened 2024-02-22
fix deploys

dyc3:


dyc3:
@coderabbitai create an issue for creating a lint to make sure the start scripts in `package.json` reference valid paths

Issue #1339, opened 2024-02-16
balancer: fix message parsing error -- `invalid type: string "true"`


dyc3:
### Current Behavior

Started seeing this log message from the balancer

```
2024-02-16T14:03:12.261509Z ERROR ott_balancer::balancer: failed to dispatch monolith message e7768db2-da5a-4726-9c07-c98392daf866: invalid type: string "true", expected a boolean at line 1 column 116
```

### Expected Behavior

All messages from monoliths should be valid json, and the balancer should be able to parse them.

### Steps To Reproduce

1. `docker-compose -f docker/with-balancer.docker-compose.yml up -d --scale opentogethertube=10`
2. Watch logs with `docker-compose -f docker/with-balancer.docker-compose.yml logs -f --tail 1`
3. See log message

### Environment

- [ ] This happens on the official site, opentogethertube.com
- [ ] This happens using a self-hosted version.
- [ ] I'm using the docker image.

### Anything else?

I think this has something to do with the `M2BGossip` mesage

Victor-M-Giraldo:
I would like to work on this.

dyc3:
I was able to reproduce this again using the `average-load` load test. Got this log message:
```
2024-02-23T03:54:39.984832Z ERROR ott_balancer::balancer: failed to dispatch monolith message cfb65b5a-7c1d-479e-9688-eef961984e7e: payload.rooms[0].room.isTemporary: invalid type: string "true", expected a boolean at line 1 column 116
```
I'm more inclined to think that there is something wrong with the monolith than the balancer.

Victor-M-Giraldo:
I have docker working both on my desktop and laptop. Can confirm I'm getting the same error. To me, this isn't really indicative of where the error might be in the code base. How can I troubleshoot more to try and track it down?

Victor-M-Giraldo:
Never mind. Can't recreate it without k6 now. Do I have to run the entire load test or is there a way to slow down the logs in the terminal to better see whats going on? I'll also start trying to mess around with the debugger.

dyc3:
you can reduce the number of vus

dyc3:
I've found an easier repro:

1. Start 1 monolith 1 balancer
2. Get an auth token from `GET /api/auth/grant`
3. Create a room via HTTP request: `POST /api/room/create`
With header:
```
Authorization: Bearer $TOKEN
```
with body
```json
{
	"name": "foobar",
	"isTemporary": "true"
}
```

[...]

PR #1384, opened 2024-02-24
fix k6 load tests so post bodies will be json

dyc3:
By default, k6/http will encode post requests as `x-www-form-urlencoded`. This can result in boolean values getting interpreted by the monolith as strings.

Related: #1339


dyc3:
@coderabbitai Create an issue to start using zod to parse and validate HTTP requests

PR #1391, opened 2024-02-24
Use zod for validation on create endpoint

### LINKED ISSUE #1339: balancer: fix message parsing error -- `invalid type: string "true"`

Victor-M-Giraldo:
related to #1339


Victor-M-Giraldo:
I think this works. Zod can also handle validation on the lengths and stuff and there is a library called zod-validation-error which can give nicer errors from what I was reading. 

Let me know if you want any changes/if I missed anything.

server/api/room.ts
```diff
@@ -543,6 +552,7 @@ router.post("/generate", async (req, res, next) => {
 
 router.post("/create", async (req, res, next) => {
 	try {
+		const _validatedBody = createRoomSchema.parse(req.body);
```
dyc3, line 555: This should be done in `createRoom`.

dyc3 (changes requested):
How does this make sure that `createRoomSchema` stays in sync with the type definition `OttApiRequestRoomCreate`? Failing to do so will cause problems in the future.

And make the PR title more descriptive.

Victor-M-Giraldo:
I'll move as much of the validation as I can to Zod. So `OttApiRequestRoomCreate` is like a schema for the request body in the create endpoint? I can see why the two being out of sync would be a problem but I'm not really sure how to account for that. I'll try and find a fix, but I might need more help if I can't.

dyc3:
I remember seeing a `zod.infer` somewhere in the docs.

It might be a good idea to move the zod schemas to ott-common instead, depending on how badly it impacts the bundle size. Vitest also has [a thing to test types](https://vitest.dev/guide/testing-types) that would also work.

Victor-M-Giraldo:
`zod.infer` returns an equivalent type definition from the defined schema. That sounds like what you want I think.

If you find it necessary I can move the zod schemas to ott-common, just give me the confirmation.

I will also take a look at Vitest testing types thing.

dyc3:
Sounds good, but I want to see how badly it affects the bundle size when you build `ott-client` because that affects site performance.

Victor-M-Giraldo:
Is there any difference in functionality if `OttApiRequestRoomCreate` is a type rather than an interface? `z.infer` gives you the type, but is it functionally the same as an interface?

dyc3:
Seems like they are basically the same thing. https://stackoverflow.com/questions/37233735/interfaces-vs-types-in-typescript

Victor-M-Giraldo:
Lint is going to fail here. Adding anything to `common/models/rest-api.ts` makes it fail.

Victor-M-Giraldo:
Little stumped on why the error handler isn't catch the zod error. Also, even when I wrap the schema parsing in a try catch, the request is succeeding in the tests and I can't for the life of me find out why.

If you could push me in the right direction/not straight up give me the answer that would be great.

server/api/room.ts
```diff
@@ -510,6 +492,15 @@ const errorHandler: ErrorRequestHandler = (err: Error, req, res) => {
 					message: err.message,
 				},
 			});
+		} else if (err instanceof ZodError) {
```
dyc3, line 495: Pay close attention to the checks being made and where in the control flow they are being made. This check will never get hit where it is now.

server/api/room.ts
```diff
@@ -119,6 +121,8 @@ const createRoom: RequestHandler<
 	OttResponseBody<OttApiResponseRoomCreate>,
 	OttApiRequestRoomCreate
 > = async (req, res) => {
+	const validatedBody = createRoomSchema.parse(req.body);
```
dyc3, line 124: Also `validatedBody` is currently not used. You should replace usages of `req.body` with the validated body object.

nit: rename `validatedBody` to `body`

Victor-M-Giraldo: Ah yeah I had it like that before but wasn't sure if it was the same. I'll do that.

server/api/room.ts
Victor-M-Giraldo: Ah yep, got it.

Victor-M-Giraldo:
- Couldn't import reserved room names into `zod-schemas.ts` because of lint errors so I just defined them there and got rid of the old ones.
- I got rid of the error in the async function since we only want to check the name, but now whenever I run the tests I get a bunch of weird output.

dyc3 (changes requested):
Looks good. Just a couple of easy fixes.

Victor-M-Giraldo:
Why do `description` and `queueMode` need to be in this schema? They weren't in the interface for the create endpoint.

dyc3 (changes requested):

Victor-M-Giraldo:
Ah okay, makes sense. 

Forgot to remove those, will do that now.

dyc3 (approved):

PR #1411, opened 2024-02-27
make balancer ids constant

### LINKED ISSUE #1338: Make balancer IDs constant from the perspective of monoliths

Currently, Balancer IDs are generated by the Monolith. When looking at log output of multiple monoliths, it can get confusing to tell which Balancer is being logged about because the ID for the same Balancer will be different for each Monolith.

Balancers should generate their own ID, and convey it to the monoliths on connect.

### Additional context

This is pretty much like #1251, except the other way around.

Victor-M-Giraldo:
closes #1338


dyc3:
I also added a `BalancerId` type in #1416. You should rebase this

Victor-M-Giraldo:
Since you fixed the commit history I just have a bunch of unresolved import errors. I'm not sure if that is only on my end. 

I also just realized I can't do this inside this function since it needs to be async to await it.

dyc3 (changes requested):
The balancer id needs to be global and constant throughout the time the balancer is alive. I'm expecting something along the lines of `static x = Lazy<...>` from `once_cell` somewhere in here.

Victor-M-Giraldo:
Okay, not a clue on how to actually send this. I've been taking a look at [this](https://docs.rs/hyper-tungstenite/latest/hyper_tungstenite/struct.WebSocketStream.htmll)

dyc3:
`WebSocketStream` impls `Sink<Message>`: https://docs.rs/hyper-tungstenite/latest/hyper_tungstenite/struct.WebSocketStream.html#impl-Sink%3CMessage%3E-for-WebSocketStream%3CT%3E

We already send messages on `WebSocketStream`s. Use those as examples.

Victor-M-Giraldo:
Okay, will take a look now.

Victor-M-Giraldo:
Know this one is going to fail as well but I feel like this is closer. Ideally I want to send it in that loop right after what I wrote. I was trying to do have another `stream.next()` but I can't have multiple mutable borrows. Do you think that is the right direction?

Victor-M-Giraldo:
`balancer stderr: console subscriber server failed: tonic::transport::Error(Transport, hyper::Error(Listen, Os { code: 98, kind: AddrInUse, message: "Address already in use" }))`

I think this is the relevant part of why the tests are failing. Could you take a look?

dyc3:
`console_subscriber` failing to bind to it's port will not kill the balancer. The harness tests are just really flaky in CI.

dyc3:
The problem is with the harness tests. Since the balancer now sends a message to the monolith on connect the `m.wait_recv()` call returns immediately without waiting to receive the client join message.

Victor-M-Giraldo:
Okay, will do.

Victor-M-Giraldo:
Think this one is good to go. Only thing is the messages get double logged [See here.](https://imgur.com/3NfxqKg). I see why, but what would the logging get changed into in the `clientmanager.ts` file?

dyc3:
yes, after a rebase, force push

dyc3 (approved):
Looks good.

PR #1435, opened 2024-03-01
Add unit tests for ott-vis-datasource and ott-collector communication

### LINKED ISSUE #1409: smoke tests for communication between ott-vis-datasource and ott-collector

cjrkoa:
Fixes #1409


packages/ott-vis-datasource/src/datasource.test.ts
dyc3, line 1: It looks like you're trying to write a unit test for the datasource. This is clearly going to end up creating more code than you're testing, which in my eyes is a code smell. Like I said before, the datasource doesn't really do any processing. There's nothing really to unit test in this package.

If anything, what you should be writing to test this package is a cypress e2e test, where you can assert that it makes an outbound request. I don't want to test implementation details.

dyc3 (changes requested):

packages/ott-vis-datasource/src/datasource.test.ts
cjrkoa: @dyc3 The way I'm reading this/interpreting tone suggests that #1409 isn't a necessary task for this sprint, am I correct? (I can still write a cypress e2e test if you think it's needed)

dyc3: Lets stick to just the cypress test. It would be nice to have a smoke test. You just need to assert that it makes an outbound request.

dyc3:
closing in favor of #1558

PR #1436, opened 2024-03-01
add ability for dns_server in discovery to be customizable

### LINKED ISSUE #1337: Add config in dns discoverer to overwrite dns_server

The current solution for #1239 in /crates/ott-common/src/discovery/dns.rs does not account for the case where `self.config.dns_server` is `Some`.

Add the ability for the code to handle the case where `self.config.dns_server` is `Some` to allow overriding the DNS server. When `dns_server` is provided, we should query the dns server provided instead of using the system default.

```mermaid
flowchart TD
    check[Check dns_server]
    check -->|Some| some[use provided DNS server]
    check -->|None| none[use system default]
```

### Additional context

_No response_

cjrkoa:
Fixes #1337

### DIFF
--- a/crates/ott-common/src/discovery/dns.rs
+++ b/crates/ott-common/src/discovery/dns.rs
@@ -67,6 +67,7 @@ impl ServiceDiscoverer for DnsServiceDiscoverer {
 mod test {
     #[allow(unused_imports)]
     use crate::discovery::DnsDiscoveryConfig;
+    #[allow(unused_imports)]
     use std::net::{IpAddr, Ipv4Addr, SocketAddr};
 
     #[tokio::test]
--- a/crates/ott-common/src/discovery/dns.rs
+++ b/crates/ott-common/src/discovery/dns.rs
@@ -65,6 +65,9 @@ impl ServiceDiscoverer for DnsServiceDiscoverer {
 }
 
 mod test {
+    use crate::discovery::DnsDiscoveryConfig;
+    use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+
     #[tokio::test]
     async fn server_deserializes_correctly() {
         let dns_discovery_config = DnsDiscoveryConfig {
--- a/crates/ott-common/src/discovery/dns.rs
+++ b/crates/ott-common/src/discovery/dns.rs
@@ -1,6 +1,9 @@
 use async_trait::async_trait;
 use tracing::info;
-use trust_dns_resolver::{config::{ResolverConfig, ResolverOpts}, Resolver, TokioAsyncResolver};
+use trust_dns_resolver::{
+    config::{NameServerConfig, Protocol, ResolverConfig, ResolverOpts},
+    Resolver, TokioAsyncResolver,
+};
 
 use super::*;
 
@@ -9,7 +12,7 @@ pub struct DnsDiscoveryConfig {
     /// The port that monoliths should be listening on for load balancer connections.
     pub service_port: u16,
     /// The DNS server to query. Optional. If not provided, the system configuration will be used instead.
-    pub dns_server: Option<String>,
+    pub dns_server: Option<SocketAddr>,
     /// The A record to query. If using docker-compose, this should be the service name for the monolith.
     pub query: String,
 }
@@ -20,10 +23,10 @@ pub struct DnsServiceDiscoverer {
 
 impl DnsServiceDiscoverer {
     pub fn new(config: DnsDiscoveryConfig) -> Self {
-            info!(
-                "Creating DnsServiceDiscoverer, DNS server: {:?}",
-                config.dns_server
-            );
+        info!(
+            "Creating DnsServiceDiscoverer, DNS server: {:?}",
+            config.dns_server
+        );
         Self { config }
     }
 }
@@ -31,11 +34,18 @@ impl DnsServiceDiscoverer {
 #[async_trait]
 impl ServiceDiscoverer for DnsServiceDiscoverer {
     async fn discover(&mut self) -> anyhow::Result<Vec<ConnectionConfig>> {
-        let resolver =
-            match self.config.dns_server {
-                None => TokioAsyncResolver::tokio_from_system_conf().expect("failed to create resolver"),
-                Some(server) => TokioAsyncResolver::tokio(ResolverConfig::new().set_domain(server), ResolverOpts::default()).expect("failed to create resolver"),
-            };
+        let resolver = match self.config.dns_server {
+            None => {
+                TokioAsyncResolver::tokio_from_system_conf().expect("failed to create resolver")
+            }
+            Some(server) => {
+                let mut resolver_config = ResolverConfig::new();
+                resolver_config.add_name_server(NameServerConfig::new(server, Protocol::Udp));
+
+                TokioAsyncResolver::tokio(resolver_config, ResolverOpts::default())
+                    .expect("failed to create resolver")
+            }
+        };
 
         let lookup = resolver.ipv4_lookup(&self.config.query).await?;
         let monoliths = lookup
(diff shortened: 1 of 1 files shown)


crates/ott-common/src/discovery/dns.rs
```diff
@@ -20,10 +20,13 @@ pub struct DnsServiceDiscoverer {
...
-            "Creating DnsServiceDiscoverer, DNS server: {:?}",
-            config.dns_server
-        );
+        match config.dns_server {
+            None => info!(
+                "Creating DnsServiceDiscoverer, DNS server: {:?}",
+                config.dns_server
+            ),
+            Some(server) => (),
+        }
```
dyc3, line 29: This just makes it so that whenever `config.dns_server` is `Some`, this log will never get printed.

I've elaborated a bit more in the issue to make it more clear what the task is.

crates/ott-common/src/discovery/dns.rs
```diff
@@ -32,7 +32,10 @@ impl DnsServiceDiscoverer {
 impl ServiceDiscoverer for DnsServiceDiscoverer {
     async fn discover(&mut self) -> anyhow::Result<Vec<ConnectionConfig>> {
         let resolver =
-            TokioAsyncResolver::tokio_from_system_conf().expect("failed to create resolver");
+            match self.config.dns_server {
+                None => TokioAsyncResolver::tokio_from_system_conf().expect("failed to create resolver"),
+                Some(server) => TokioAsyncResolver::tokio(ResolverConfig::new().set_domain(server), ResolverOpts::default()).expect("failed to create resolver"),
```
dyc3, line 37: I don't think `set_domain` does what we want here. [`add_name_server`](https://docs.rs/trust-dns-resolver/latest/trust_dns_resolver/config/struct.ResolverConfig.html#method.add_name_server) is probably what you want.

You also might want to change `dns_server` from a `String` to a `SocketAddr` or `IpAddr`.

dyc3 (commented):
Looking better.

tip: you might want to turn on format on save

crates/ott-common/src/discovery/dns.rs
```diff
@@ -63,3 +63,22 @@ impl ServiceDiscoverer for DnsServiceDiscoverer {
...
+    #[tokio::test]
+    async fn server_deserializes_correctly() {
+        let dns_discovery_config = DnsDiscoveryConfig {
+            service_port: 8080,
+            dns_server: Option::Some(SocketAddr::new(
+                IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)),
+                8080,
+            )),
+            query: "".to_string(),
+        };
```
dyc3, line 82: This is just constructing the struct. Deserializing is the process of going from a format like `json` or `toml` into a native data type, like `DnsDiscoveryConfig`.

crates/ott-common/src/discovery/dns.rs
```diff
@@ -63,3 +63,22 @@ impl ServiceDiscoverer for DnsServiceDiscoverer {
         DiscoveryMode::Polling(Duration::from_secs(10))
     }
 }
+
+mod test {
+    #[tokio::test]
+    async fn server_deserializes_correctly() {
```
dyc3, line 74: This test doesn't need to be async

dyc3 (changes requested):

crates/ott-common/src/discovery/dns.rs
```diff
@@ -50,3 +63,26 @@ impl ServiceDiscoverer for DnsServiceDiscoverer {
...
+    #[allow(unused_imports)]
+    use serde_json::json;
+    #[allow(unused_imports)]
+    use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+
+    #[tokio::test]
+    async fn server_deserializes_correctly() {
+        let json = json!({
+            "service_port": 8080,
+            "dns_server": "127.0.0.1:8080",
```
dyc3, line 79: Does this mean that the port is required? Typically dns is served on port 53 and we should default to that if the user doesn't provide it.

crates/ott-common/src/discovery/dns.rs
```diff
@@ -50,3 +63,26 @@ impl ServiceDiscoverer for DnsServiceDiscoverer {
...
+            "query": "".to_string(),
+        });
+
+        let config: DnsDiscoveryConfig =
+            serde_json::from_value(json).expect("Failed to deserialize json");
+
+        assert_eq!(config.query, "".to_string())
```
dyc3, line 86: This should assert on the value of `dns_server`.

crates/ott-common/src/discovery/dns.rs
```diff
@@ -50,3 +63,26 @@ impl ServiceDiscoverer for DnsServiceDiscoverer {
...
     }
 }
+
+mod test {
+    #[allow(unused_imports)]
+    use crate::discovery::DnsDiscoveryConfig;
+    #[allow(unused_imports)]
+    use serde_json::json;
+    #[allow(unused_imports)]
+    use std::net::{IpAddr, Ipv4Addr, SocketAddr};
```
dyc3, line 73: nit

```suggestion
    use super::*;
```

crates/ott-common/src/discovery/dns.rs
```diff
@@ -50,3 +63,26 @@ impl ServiceDiscoverer for DnsServiceDiscoverer {
         DiscoveryMode::Polling(Duration::from_secs(10))
     }
 }
+
+mod test {
```
dyc3, line 67: Without this, this code will always be compiled. This makes it so that it only compiles when you run `cargo test`.
```suggestion
#[cfg(test)]
mod test {
```

dyc3 (changes requested):
Looking good. Your PR title could be a little better though (kinda missing the "what" and "why").

cjrkoa:
> Looking good. Your PR title could be a little better though (kinda missing the "what" and "why").

we can talk about the pr title during the meeting tomorrow

cjrkoa:
for the record vscode complains when I don't have the async keyword or the other imports

crates/ott-common/src/discovery/dns.rs
```diff
@@ -50,3 +63,25 @@ impl ServiceDiscoverer for DnsServiceDiscoverer {
         DiscoveryMode::Polling(Duration::from_secs(10))
     }
 }
+#[cfg(test)]
+mod test {
+    use super::*;
+    use serde_json::json;
+    use std::net::Ipv4Addr;
+
+    #[tokio::test]
```
dyc3, line 72: `tokio::test` is for async tests

```suggestion
    #[test]
```

crates/ott-common/src/discovery/dns.rs
cjrkoa: it seems like the port is required for creating a socketaddr struct: https://doc.rust-lang.org/std/net/enum.SocketAddr.html

dyc3: I would much prefer we don't require the port to be specified. Like I said before, we should default to 53 if the user doesn't provide it.

This will likely require customizing how serde deserializes this field. Take a look at:
- https://brokenco.de/2020/08/03/serde-deserialize-with-string.html
- https://serde.rs/custom-date-format.html

Keep in mind you only need to implement the deserialize side.

cjrkoa: I think this should be filed as a separate task:

1. The dns_server is customizable and this wasn't a requirement mentioned in the original issue
2. (I don't think) there was any way I could have reasonably anticipated the port needing to be specified would be a problem
3. For the sake of sprint planning
4. (imo) This just kinda feels like a separate thing

Completely understand if you don't want to merge this PR yet though. I'd be happy to hop on a call tomorrow or talk about it more on Thursday if you wanted.

dyc3: @coderabbitai can you open an issue to make the port for `dns_server` optional?

dyc3 (approved):

PR #1452, opened 2024-03-06
update balancer to accept json values for room metadata

### LINKED ISSUE #1339: balancer: fix message parsing error -- `invalid type: string "true"`

Victor-M-Giraldo:
Closes #1339

### DIFF
--- a/crates/ott-balancer-protocol/src/monolith.rs
+++ b/crates/ott-balancer-protocol/src/monolith.rs
@@ -206,17 +206,17 @@ where
 /// Metadata about a room, according to the Monolith.
 pub struct RoomMetadata {
     pub name: RoomName,
-    pub title: String,
-    pub description: String,
+    pub title: serde_json::Value,
+    pub description: serde_json::Value,
     #[serde(rename = "isTemporary")]
-    pub is_temporary: bool,
+    pub is_temporary: serde_json::Value,
     pub visibility: Visibility,
     #[serde(rename = "queueMode")]
-    pub queue_mode: String,
+    pub queue_mode: serde_json::Value,
     #[serde(rename = "currentSource")]
     pub current_source: serde_json::Value,
     /// The number of clients in this room.
-    pub users: u32,
+    pub users: serde_json::Value,
 }
 
 impl RoomMetadata {
--- a/crates/ott-balancer-protocol/src/monolith.rs
+++ b/crates/ott-balancer-protocol/src/monolith.rs
@@ -206,16 +206,19 @@ where
 /// Metadata about a room, according to the Monolith.
 pub struct RoomMetadata {
     pub name: RoomName,
+    #[typeshare(serialized_as = "unknown")]
     pub title: serde_json::Value,
+    #[typeshare(serialized_as = "unknown")]
     pub description: serde_json::Value,
-    #[serde(rename = "isTemporary")]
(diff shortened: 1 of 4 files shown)


Victor-M-Giraldo:
Didn't make name a `serde_json::Value` since I wasn't sure how I could convert it into a `RoomName` with the pointers and stuff.

dyc3 (approved):
Looks good.

Victor-M-Giraldo:
Do you know why the lint is failing? I run yarn lint on my end and it goes through fine.

server/generated.ts
```diff
@@ -51,14 +51,14 @@ export enum Visibility {
 /** Metadata about a room, according to the Monolith. */
 export interface RoomMetadata {
 	name: RoomName;
-	title: string;
-	description: string;
-	isTemporary: boolean;
+	title: unknown;
+	description: unknown;
+	isTemporary: unknown;
```
dyc3, line 56: Did you manually change these to unknown? Typeshare doesn't treat these as unknown by default, and it doesn't have an escape hatch to set the type manually. The codegen script has a hack to manually tweak the output of typeshare so it works a little better for us (that's why `currentSource` wasn't causing problems)

Victor-M-Giraldo: Yes, I ran the typeshare script and these didn't get updated (or well, they did but to `Value`)

dyc3: The CI job enforces that running the codegen script produces the exact same output as what is in `generated.ts`

crates/ott-balancer-protocol/src/monolith.rs
```diff
@@ -206,16 +206,19 @@ where
...
+    #[typeshare(serialized_as = "unknown")]
     pub description: serde_json::Value,
-    #[serde(rename = "isTemporary")]
+    #[typeshare(serialized_as = "unknown")]
     pub is_temporary: serde_json::Value,
     pub visibility: Visibility,
-    #[serde(rename = "queueMode")]
+    #[typeshare(serialized_as = "unknown")]
     pub queue_mode: serde_json::Value,
-    #[serde(rename = "currentSource")]
```
dyc3, line 216: these renames need to stay. you can put multiple attributes on a field

dyc3 (changes requested):

crates/ott-balancer-protocol/src/monolith.rs
Victor-M-Giraldo: ah, makes sense. thats why the names changed on the monolith end. will change that in a second.

dyc3 (approved):

dyc3 (approved):

PR #1455, opened 2024-03-06
use purgecss to remove tons of unused css from bundles

dyc3:
not entirely sure I want to merge this considering how much hacky stuff I had to do

I think it would be a good idea to land some snapshot tests first to make sure there aren't any differences

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -42,6 +42,7 @@
   },
   "devDependencies": {
     "@cypress/vue": "^5.0.3",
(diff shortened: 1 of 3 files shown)


dyc3:
There's still some missing styles that need to be resolved before this can be merged

- [ ] hero thing on landing page not centered
- [ ] browse rooms empty list thing not centered
- [ ] empty video queue thing not centered
- [ ] footer on landing page not centered
- [ ] footer on room view not centered

PR #1476, opened 2024-03-11
Switch to using zod for validation on voting API endpoints

Victor-M-Giraldo:
Related #1385

### DIFF
--- a/common/models/zod-schemas.ts
+++ b/common/models/zod-schemas.ts
@@ -6,7 +6,7 @@ import { string, z } from "zod";
 // These strings are not allowed to be used as room names.
 const RESERVED_ROOM_NAMES = ["list", "create", "generate"];
 
-export const createRoomSchema = z.object({
+export const OttApiRequestRoomCreateSchema = z.object({
 	name: z
 		.string()
 		.min(3, "too short, must be atleast 3 characters")
@@ -25,6 +25,6 @@ const VideoIdSchema = z.object({
 	id: z.string(),
 });
 
-export const voteSchema = z.object({
+export const OttApiRequestVoteSchema = z.object({
 	...VideoIdSchema.shape,
 });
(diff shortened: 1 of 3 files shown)


Victor-M-Giraldo:
Couple of things.
- Do you want unit tests for these?
- I run tests on my end and I get a bunch of weird output. Been happening for a while and I've tried finding out why on Google/Stack and no clue. [See here](https://i.imgur.com/WfryHVS.png)

common/models/zod-schemas.ts
```diff
@@ -18,3 +19,9 @@ export const createRoomSchema = z.object({
 	visibility: z.nativeEnum(Visibility).default(Visibility.Public).optional(),
 	queueMode: z.nativeEnum(QueueMode).optional(),
 });
+
+export const voteSchema = z.object({
+	service: z.enum(ALL_VIDEO_SERVICES),
+	id: z.string(),
```
dyc3, line 25: You should create a schema for `VideoId`, it gets used quite a lot and iirc you can use schemas in schemas

server/api/room.ts
```diff
@@ -304,17 +304,9 @@ const undoEvent: RequestHandler<{ name: string }> = async (req, res) => {
 };
 
 const addVote: RequestHandler<{ name: string }, unknown, OttApiRequestVote> = async (req, res) => {
-	if (!req.token) {
```
dyc3, line 307: `req.token` is not on the body. You shouldn't be touching auth tokens with these validation PRs

dyc3 (changes requested):

common/models/zod-schemas.ts
Victor-M-Giraldo: Yea, I realized that when I was getting to work on the queue API endpoints yesterday because it would make the union a lot easier.

dyc3:
Is this ready?

Victor-M-Giraldo:
Think so.

dyc3 (changes requested):
Looks good, just a couple minor changes

dyc3 (approved):

PR #1477, opened 2024-03-11
Switch to using zod for validation on queue API endpoints

Victor-M-Giraldo:
Related #1385

### DIFF
--- a/server/api/room.ts
+++ b/server/api/room.ts
@@ -32,7 +32,7 @@ import { getApiKey } from "../admin";
 import { v4 as uuidv4 } from "uuid";
 import { counterHttpErrors } from "../metrics";
 import { conf } from "../ott-config";
-import { createRoomSchema, voteSchema } from "ott-common/models/zod-schemas";
+import { OttApiRequestRoomCreateSchema, OttApiRequestVoteSchema } from "ott-common/models/zod-schemas";
 import { ZodError } from "zod";
 import { fromZodError } from "zod-validation-error";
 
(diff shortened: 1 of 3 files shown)


Victor-M-Giraldo:
For these two I guess there isn't a lot of validation going on? Can't I just get rid of the errors being thrown at the end, since in the case that these are missing Zod will handle throwing the error?

dyc3:
No these endpoints aren't very complicated validation wise.

server/api/room.ts
```diff
@@ -355,67 +351,68 @@ const addToQueue: RequestHandler<
...
+		points = 3 * body.videos.length;
 	}
 	if (!(await consumeRateLimitPoints(res, req.ip, points))) {
 		return;
 	}
 	const room = (await roommanager.getRoom(req.params.name)).unwrap();
 
-	let roomRequest: AddRequest;
-	if ("videos" in req.body) {
+	let roomRequest: AddRequest | undefined;
```
dyc3, line 364: This type needs to remain the same so that the type checking guarentees this will be set.

server/api/room.ts
```diff
@@ -355,67 +351,68 @@ const addToQueue: RequestHandler<
...
-	res.json({
-		success: true,
-	});
+
+	if (roomRequest !== undefined) {
+		await room.processUnauthorizedRequest(roomRequest, { token: req.token! });
+		res.json({
+			success: true,
+		});
+	}
```
dyc3, line 390: Which would make this check for undefined unnecessary.

The way this is now if there is a logic error and roomRequest is undefined, it will cause the server to never send a response.

dyc3 (changes requested):
Gonna need to rebase this one

Victor-M-Giraldo:
> Gonna need to rebase this one

Going to hold off on that until Tuesday. Don't want to accidentally merge again even though I'm pretty sure I know how to do it now.

PR #1486, opened 2024-03-12
ci: allow people to rerun failed jobs by commenting `/rerun`

dyc3:


dyc3:
it appears this can't exactly be tested before it gets merged.

dyc3:
It's not working because it's missing permissions, but I'm pretty sure I've granted the right permissions. This is something I will have to work on tomorrow.

PR #1534, opened 2024-03-19
Switch to using zod for validation on queue API endpoints 

Victor-M-Giraldo:
- Related #1385

### DIFF
--- a/server/api/room.ts
+++ b/server/api/room.ts
@@ -35,6 +35,8 @@ import { conf } from "../ott-config";
 import {
 	OttApiRequestRoomCreateSchema,
 	OttApiRequestVoteSchema,
+	OttApiRequestAddToQueueSchema,
+	OttApiRequestRemoveFromQueueSchema,
 } from "ott-common/models/zod-schemas";
 import { ZodError } from "zod";
 import { fromZodError } from "zod-validation-error";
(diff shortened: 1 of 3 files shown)


server/api/room.ts
```diff
@@ -350,37 +352,37 @@ const addToQueue: RequestHandler<
...
+		points = 3 * body.videos.length;
 	}
 	if (!(await consumeRateLimitPoints(res, req.ip, points))) {
 		return;
 	}
 	const room = (await roommanager.getRoom(req.params.name)).unwrap();
 
-	let roomRequest: AddRequest;
-	if ("videos" in req.body) {
+	let roomRequest: AddRequest | undefined;
```
dyc3, line 365: I would prefer this to remain as just `AddRequest`. Do we need to add undefined?

server/api/room.ts
```diff
@@ -392,25 +394,24 @@ const removeFromQueue: RequestHandler<
...
 > = async (req, res) => {
+	const body = OttApiRequestRemoveFromQueueSchema.parse(req.body);
 	let points = 5;
 	if (!(await consumeRateLimitPoints(res, req.ip, points))) {
 		return;
 	}
 	const room = (await roommanager.getRoom(req.params.name)).unwrap();
 
-	if (req.body.service && req.body.id) {
+	if (body.service && body.id) {
```
dyc3, line 404: I don't think we need this check anymore

dyc3 (changes requested):

server/api/room.ts
Victor-M-Giraldo: No, that's my bad. I just forgot to delete that.

dyc3 (approved):

PR #1542, opened 2024-03-21
Changed the path to the correct environment location

cloudmaker97:
By creating a container of this project with the details in the documentation, the path is not mounted correctly, instead i updated it to the correct path, so that not docker-aware users are able to run this with ease.


dyc3 (approved):
Whoops! Good catch. Thank you!

PR #1555, opened 2024-03-22
Switch to using zod for validation for account recovery endpoints.

Victor-M-Giraldo:
- Related #1385

### DIFF
--- a/common/models/zod-schemas.ts
+++ b/common/models/zod-schemas.ts
@@ -42,7 +42,6 @@ export const OttApiRequestRemoveFromQueueSchema = z.object({
 	...VideoIdSchema.shape,
 });
 
-
 export const OttApiRequestAccountRecoveryStartSchema = z.object({
 	email: z.string().email().optional(),
 	username: z.string().optional(),
--- a/common/models/zod-schemas.ts
+++ b/common/models/zod-schemas.ts
@@ -41,3 +41,14 @@ export const OttApiRequestAddToQueueSchema = z.union([
 export const OttApiRequestRemoveFromQueueSchema = z.object({
 	...VideoIdSchema.shape,
 });
+
+
+export const OttApiRequestAccountRecoveryStartSchema = z.object({
+	email: z.string().email().optional(),
+	username: z.string().optional(),
+});
+
+export const OttApiRequestAccountRecoveryVerifySchema = z.object({
+	verifyKey: z.string(),
+	newPassword: z.string(),
+});
--- a/common/models/zod-schemas.ts
+++ b/common/models/zod-schemas.ts
@@ -44,7 +44,7 @@ export const OttApiRequestRemoveFromQueueSchema = z.object({
 
 export const OttApiRequestAccountRecoveryStartSchema = z.union([
 	z.object({
-		email: z.string().email(),
+		email: z.string().email().min(3),
 	}),
 	z.object({
 		username: z.string(),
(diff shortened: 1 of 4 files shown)


Victor-M-Giraldo:
Still need to fix the tests when I get home later.

common/models/zod-schemas.ts
```diff
@@ -41,3 +41,13 @@ export const OttApiRequestAddToQueueSchema = z.union([
 export const OttApiRequestRemoveFromQueueSchema = z.object({
 	...VideoIdSchema.shape,
 });
+
+export const OttApiRequestAccountRecoveryStartSchema = z.object({
+	email: z.string().email().optional(),
+	username: z.string().optional(),
```
dyc3, line 47: This needs to be one or the other. At least one of these is required.

dyc3 (changes requested):

Victor-M-Giraldo:
Something like this?

server/usermanager.ts
```diff
@@ -744,28 +750,20 @@ const accountRecoveryStart: RequestHandler<
...
-	if (!("email" in req.body) && !("username" in req.body)) {
-		throw new BadApiArgumentException(
-			"email,username",
-			"Either email or username is required."
-		);
-	}
-
 	const query = {
-		user: req.body.email ?? req.body.username,
+		user: (body as { email?: string }).email ?? (body as { username?: string })?.username,
```
dyc3, line 759: While this technically works, it wont throw type errors if the type of `body` changes because `as` forces a type override. Take a look at the type narrowing docs: https://www.typescriptlang.org/docs/handbook/2/narrowing.html

server/usermanager.ts
```diff
@@ -744,28 +750,20 @@ const accountRecoveryStart: RequestHandler<
...
 	const user = await getUser(query);
 
-	if (!user.email) {
-		throw new NoEmail();
+	if (user.email) {
+		(await sendPasswordResetEmail(user.email)).unwrap();
 	}
 
-	(await sendPasswordResetEmail(user.email)).unwrap();
-
```
dyc3, line 768: Try not to change the functionality of these endpoints. This makes the endpoint fake succeed when there is no email on the account.

This logic needs to remain as is because its possible for an account to have no email associated with it (like if they just use a social login).

dyc3 (changes requested):
Yeah, better.

Victor-M-Giraldo:
Something like this?

common/models/zod-schemas.ts
```diff
@@ -41,3 +41,17 @@ export const OttApiRequestAddToQueueSchema = z.union([
 export const OttApiRequestRemoveFromQueueSchema = z.object({
 	...VideoIdSchema.shape,
 });
+
+export const OttApiRequestAccountRecoveryStartSchema = z.union([
+	z.object({
+		email: z.string().email(),
```
dyc3, line 47: The test could be failing because technically we didn't validate this field as an email before. We probably should though. Its possible i used an email that zod thinks is invalid in the test.

dyc3 (commented):
Yeah, something like this.

common/models/zod-schemas.ts
dyc3, line 47: if we change this validation to `.email().min(3)` can we avoid the unit test changes? Ideally those shouldn't need to change because 3 character email addresses like `a@b` _are_ valid emails, and it's not like we require email validation on signup.

Victor-M-Giraldo: We need to tack on `.com` or something similar, but other than that, yes.
See [here](https://colinhacks.com/essays/reasonable-email-regex).

dyc3: Alright, I'm fine with this then.

dyc3 (approved):

PR #1564, opened 2024-03-27
add deserialization to SocketAddr struct to make port for dns_config optional

### LINKED ISSUE #1450: Make port for `dns_server` optional in `DnsDiscoveryConfig`
In the context of PR https://github.com/dyc3/opentogethertube/pull/1436, it was discussed that the `dns_server` field in the `DnsDiscoveryConfig` struct should allow the port to be optional, defaulting to 53 if not provided. This change aims to simplify configuration by not requiring users to specify a port when it's unnecessary.

This issue was requested to be filed by @dyc3 in a comment: https://github.com/dyc3/opentogethertube/pull/1436#discussion_r1511851671

The desired behavior is for the deserialization of the `dns_server` field to support omitting the port, and if omitted, the system should default to using port 53. This may require customizing the deserialization process, potentially using Serde's `deserialize_with` attribute or similar mechanisms.

cjrkoa:
fixes #1450

### DIFF
--- a/crates/ott-common/src/discovery/dns.rs
+++ b/crates/ott-common/src/discovery/dns.rs
@@ -5,7 +5,7 @@ use hickory_resolver::{
     config::{NameServerConfig, Protocol, ResolverConfig, ResolverOpts},
     TokioAsyncResolver,
 };
-use serde::{Deserializer, Serialize};
+use serde::Deserializer;
 use tracing::info;
 
 use super::*;
@@ -37,10 +37,10 @@ impl FromValue for DnsDiscoveryConfig {
                 "dns_server".to_string(),
                 serde_json::Value::from_str("").unwrap(),
             );
-            let value = value.as_object_mut().insert(&mut kv);
+            let _value = value.as_object_mut().insert(&mut kv);
         }
 
-        return serde_json::from_value(value);
+        serde_json::from_value(value)
     }
 }
 
@@ -155,6 +155,6 @@ mod test {
         let config: DnsDiscoveryConfig =
             DnsDiscoveryConfig::from_value(json).expect("Failed to deserialize json");
 
-        assert_eq!(config.dns_server.is_some(), true);
+        assert_eq!(config.dns_server.is_some());
     }
 }
--- a/crates/ott-common/src/discovery/dns.rs
+++ b/crates/ott-common/src/discovery/dns.rs
@@ -32,10 +32,6 @@ where
 {
     let buf = String::deserialize(deserializer)?;
 
-    if buf.is_empty() {
-        return Ok(None);
-    }
-
     match IpAddr::from_str(&buf) {
         Ok(ip) => Ok(Some(SocketAddr::new(ip, 53))),
         Err(_) => match SocketAddr::from_str(&buf) {
--- a/crates/ott-common/src/discovery/dns.rs
+++ b/crates/ott-common/src/discovery/dns.rs
@@ -36,9 +36,9 @@ where
         _ => (),
     }
 
-    return Ok(Some(
+    Ok(Some(
         SocketAddr::from_str(&buf).expect("Error: Could not unwrap"),
-    ));
+    ))
 }
 
 pub struct DnsServiceDiscoverer {
--- a/crates/ott-common/src/discovery/dns.rs
+++ b/crates/ott-common/src/discovery/dns.rs
@@ -31,7 +31,7 @@ where
 {
     let mut buf = String::deserialize(deserializer)?;
 
-    if Option::is_none(buf.find(':')) {
+    if Option::is_none(&buf.find(':')) {
         buf.push_str(":53");
     }
 
--- a/crates/ott-common/src/discovery/dns.rs
+++ b/crates/ott-common/src/discovery/dns.rs
@@ -128,7 +128,6 @@ mod test {
     fn dns_server_is_optional() {
         let json = json!({
             "service_port": 8080,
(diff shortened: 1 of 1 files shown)


crates/ott-common/src/discovery/dns.rs
```diff
@@ -21,6 +25,21 @@ pub struct DnsDiscoveryConfig {
...
+    D: Deserializer<'de>,
+{
+    let mut buf = String::deserialize(deserializer)?;
+
+    if Option::is_none(&buf.find(':')) {
+        buf.push_str(":53");
+    }
+
+    Ok(Some(
+        SocketAddr::from_str(&buf).expect("Error: Could not unwrap"),
```
dyc3, line 39: You really shouldn't be calling expect here anyway, you are already returning a `Result`. `expect` triggers a panic, which is not practically recoverable. Panicking is reserved for instances where there is absolutely no possible way for the application to continue, and parsing failures do not fall into that category.

nit: "Could not unwrap" is too generic. Give context in error messages so they are actionable or searchable. Could not unwrap what? Could not parse what? Consider the user reading the error message.

dyc3 (changes requested):
It's about 90% there.

crates/ott-common/src/discovery/dns.rs
```diff
@@ -29,15 +29,17 @@ fn deserialize_dns_server<'de, D>(deserializer: D) -> Result<Option<SocketAddr>,
...
-        buf.push_str(":53");
+    if buf.is_empty() {
+        return Ok(None);
     }
 
-    Ok(Some(
-        SocketAddr::from_str(&buf).expect("Error: Could not unwrap"),
-    ))
+    if IpAddr::from_str(&buf).is_ok() {
+        Ok(Some(SocketAddr::new(IpAddr::from_str(&buf).unwrap(), 53)))
```
dyc3, line 39: `unwrap` panics, and you're repeating the `from_str` operation. use an `if let ...` or `match` instead

crates/ott-common/src/discovery/dns.rs
```diff
@@ -29,15 +29,17 @@ fn deserialize_dns_server<'de, D>(deserializer: D) -> Result<Option<SocketAddr>,
...
+    } else {
+        Ok(Some(SocketAddr::from_str(&buf).unwrap()))
```
dyc3, line 41: `unwrap` panics

crates/ott-common/src/discovery/dns.rs
```diff
@@ -107,4 +123,18 @@ mod test {
 
         assert_eq!(config.dns_server, Some(([127, 0, 0, 1], 53).into()));
     }
+
+    #[test]
+    fn dns_server_is_optional() {
+        let json = json!({
+            "service_port": 8080,
+            "dns_server": "",
```
dyc3, line 131: This is the case I'm more concerned about. Here you are still providing `dns_server`. This change will verify `dns_server` is optional.

```suggestion
```

dyc3 (changes requested):

crates/ott-common/src/discovery/dns.rs
```diff
@@ -21,6 +25,44 @@ pub struct DnsDiscoveryConfig {
...
+        return Ok(None);
+    }
+
+    match IpAddr::from_str(&buf) {
+        Ok(ip) => Ok(Some(SocketAddr::new(ip, 53))),
+        Err(_) => match SocketAddr::from_str(&buf) {
+            Ok(socket) => Ok(Some(socket)),
+            Err(_) => Ok(None),
+        },
+    }
```
dyc3, line 63: This looks good. Great work. The only problem is that it should return an error when all parsing fails instead of ignoring the value. We _do_ want to surface those errors to the user.

crates/ott-common/src/discovery/dns.rs
```diff
@@ -21,6 +25,44 @@ pub struct DnsDiscoveryConfig {
...
+            kv.insert(
+                "dns_server".to_string(),
+                serde_json::Value::from_str("").unwrap(),
+            );
+            let _value = value.as_object_mut().insert(&mut kv);
+        }
+
+        serde_json::from_value(value)
+    }
+}
```
dyc3, line 45: Modifying the config is a bad idea. It's supposed to be immutable. Also, this doesn't actually change the way `DnsDiscoveryConfig` is deserialized.

You want [`#[serde(default)]`](https://serde.rs/attr-default.html) (see `polling_interval`)

dyc3 (changes requested):

cjrkoa:
I think I'll figure out the error thing tomorrow morning

crates/ott-common/src/discovery/dns.rs
```diff
@@ -88,4 +116,46 @@ mod test {
...
+            "service_port": 8080,
+            "query": "".to_string(),
+        });
+
+        let config: DnsDiscoveryConfig =
+            serde_json::from_value(json).expect("Failed to deserialize json");
+
+        assert_eq!(
+            config.dns_server,
+            Some(SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 53))
```
dyc3, line 146: This scenario should result in `dns_server` being `None`.

dyc3 (changes requested):

dyc3 (approved):
Looks good.

cjrkoa:
never mind on the re-review

PR #1572, opened 2024-03-27
client: refactor media controls to use composables instead of hacky component refs

dyc3:


dyc3:
After manually testing some more and passing the new e2e tests, I'm feeling pretty confident about this code.

PR #1573, opened 2024-03-27
Make load balancing configurable

### LINKED ISSUE #1294: make the load balancing algorithm configurable

We want to be able to swap out load balancing algorithms without recompiling.

add an enum of some kind to `BalancerConfig` to make the `MonolithSelection` that is used by the `BalancerContext` configurable.

### Additional context

_No response_

Victor-M-Giraldo:
- Closes #1294


Victor-M-Giraldo:
```
impl Default for BalancerContext {
    fn default() -> Self {
        BalancerContext {
            clients: HashMap::default(),
            monoliths: HashMap::default(),
            rooms_to_monoliths: HashMap::default(),
            monoliths_by_region: HashMap::default(),
            monolith_selection: Box::new(BalancerConfig::get().selection_strategy),
        }
    }
}
```

Shouldn't something like this work? It's asking me to implement the trait for the enum though I think which makes no sense to me. Am I going in the wrong direction here though?

dyc3:
It only works if you implement the trait for the enum. The enum does not automatically implement traits that the variants implement. The enum is a completely new type.

crates/ott-balancer/src/balancer.rs
```diff
@@ -191,7 +191,7 @@ pub struct BalancerContext {
     pub monoliths: HashMap<MonolithId, BalancerMonolith>,
     pub rooms_to_monoliths: HashMap<RoomName, RoomLocator>,
     pub monoliths_by_region: HashMap<String, Vec<MonolithId>>,
-    pub monolith_selection: Box<dyn MonolithSelection + Send + Sync + 'static>,
+    pub monolith_selection: Box<dyn MonolithSelection + 'static + Send + Sync>,
```
dyc3, line 194: `enum_dispatch` allows us to drop the `Box<dyn T>` and replace it with the enum type

dyc3 (changes requested):

crates/ott-balancer/src/balancer.rs
```diff
@@ -202,7 +202,7 @@ impl Default for BalancerContext {
             monoliths: HashMap::default(),
             rooms_to_monoliths: HashMap::default(),
             monoliths_by_region: HashMap::default(),
-            monolith_selection: Box::<MinRoomsSelector>::default(),
+            monolith_selection: BalancerConfig::get().selection_strategy,
```
dyc3, line 205: This shouldn't be done in the `Default` impl. `Default` should return the exact same value every time, eg the "default" value. I would recommend deriving `Default` for `MonolithSelectionStrategy`. `BalancerConfig::get()` can panic if the config is not initialized (which is why the ci is failing)

Victor-M-Giraldo:
I'm trying to figure out why these tests are failing. I'm running `cargo test` with `RUST_BACKTRACE = 1`. Do you think you could look?

dyc3 (changes requested):

crates/ott-balancer/src/balancer.rs
dyc3: See: https://doc.rust-lang.org/std/default/trait.Default.html#enums

Victor-M-Giraldo: Okay, makes sense

crates/ott-balancer/src/config.rs
```diff
@@ -19,6 +19,11 @@ pub enum MonolithSelectionStrategy {
     MinRooms(MinRoomsSelector),
 }
 
+impl Default for MonolithSelectionStrategy {
+    fn default() -> Self {
+        MonolithSelectionStrategy::MinRooms(MinRoomsSelector)
+    }
+}
```
dyc3, line 26: You should be able to `#[derive(Default)]` and select which enum variant is the default by tagging it with the `#[default]` attribute.

dyc3 (commented):
Looking good

crates/ott-balancer/src/config.rs
Victor-M-Giraldo: I tried to do that, but compiler gave me the error that I can't do that since it's not a unit variant or something along the lines.

dyc3: Ah ok. Don't worry about it then.

crates/ott-balancer/Cargo.toml
```diff
@@ -35,6 +35,7 @@ route-recognizer = "0.3.1"
 once_cell.workspace = true
 pin-project.workspace = true
 prometheus.workspace = true
+enum_dispatch = "0.3.13"
```
dyc3, line 38: This should be in the root `Cargo.toml` and referenced in here just like all the other dependencies here.

dyc3 (changes requested):

crates/ott-balancer/src/balancer.rs
```diff
@@ -186,30 +186,26 @@ impl BalancerLink {
...
+        if BalancerConfig::get().selection_strategy.is_some() {
+            BalancerContext {
+                monolith_selection: BalancerConfig::get()
+                    .selection_strategy
+                    .expect("selection strategy not set"),
+                ..BalancerContext::default()
+            }
+        } else {
+            BalancerContext::default()
+        }
```
dyc3, line 208: do this in `lib.rs` after the config gets loaded

Victor-M-Giraldo:
Why is the config failing to be initialized. I thought this logic is okay. If `Some(selection_strategy)` set it in the balancer context, else use the default, where the default for `selection_strategy` would be `Min_rooms`

crates/ott-balancer/src/lib.rs
```diff
@@ -57,6 +57,19 @@ pub async fn run() -> anyhow::Result<()> {
 
     let config = BalancerConfig::get();
 
+    if config.selection_strategy.is_some() {
+        info!("Using selection strategy: {:?}", config.selection_strategy);
+        BalancerContext {
+            monolith_selection: config
+                .selection_strategy
+                .expect("selection strategy not set"),
```
dyc3, line 65: You should use `if let Some(...) = ...` here instead of calling `expect`

crates/ott-balancer/src/lib.rs
```diff
@@ -57,6 +57,19 @@ pub async fn run() -> anyhow::Result<()> {
...
+            ..BalancerContext::default()
+        }
+    } else {
+        info!("Using default selection strategy");
+        BalancerContext::default()
+    };
```
dyc3, line 71: This `BalancerContext` is not actually being used, and this needs to happen after the `tracing_subscriber::registry().init()` otherwise these log lines will never appear, so it needs to go after the "Loaded config" log line

dyc3 (changes requested):

crates/ott-balancer/src/lib.rs
Victor-M-Giraldo: Just to clarify before I make this change, the config isn't fully loaded until here `info!("Loaded config: {:?}", config);` because it is being manipulated in some way. `BalancerConfig::load` just allows it to be manipulated from the file path.

dyc3: No the config is loaded at `BalancerConfig::load()`. The `info!` log macros don't work until tracing subscriber is initialized.

dyc3 (approved):

PR #1579, opened 2024-03-28
Switch to using zod for validation for patch endpoint.

Victor-M-Giraldo:
- Related #1385

### DIFF
--- a/common/models/zod-schemas.ts
+++ b/common/models/zod-schemas.ts
@@ -68,12 +68,8 @@ const CategorySchema = z.enum([
 ]);
 
 const GrantSchema = z.tuple([
-	z.object({
-		role: z.nativeEnum(Role),
-	}),
-	z.object({
-		grantMask: z.number(),
-	}),
+	z.nativeEnum(Role),
+	z.number(),
 ]);
 
 const RoomSettingsSchema = z.object({
@@ -89,7 +85,6 @@ const RoomSettingsSchema = z.object({
 
 const ClaimSchema = z.object({
 	claim: z.boolean().optional(),
-	...RoomSettingsSchema.shape,
 });
 
 export const OttApiRequestPatchRoomSchema = z.union([RoomSettingsSchema, ClaimSchema]);
--- a/common/models/zod-schemas.ts
+++ b/common/models/zod-schemas.ts
@@ -67,10 +67,7 @@ const CategorySchema = z.enum([
 	"preview",
 ]);
 
-const GrantSchema = z.tuple([
-	z.nativeEnum(Role),
-	z.number(),
-]);
+const GrantSchema = z.tuple([z.nativeEnum(Role), z.number()]);
 
 const RoomSettingsSchema = z.object({
 	title: z.string().max(255).optional(),
--- a/common/models/zod-schemas.ts
+++ b/common/models/zod-schemas.ts
@@ -69,8 +69,8 @@ const CategorySchema = z.enum([
 
 const GrantSchema = z.tuple([z.nativeEnum(Role), z.number()]);
 
-const RoomSettingsSchema = z.object({
-	title: z.string().max(255).optional(),
+export const RoomSettingsSchema = z.object({
+	title: z.string().max(254).optional(),
 	description: z.string().optional(),
 	visibility: z.nativeEnum(Visibility).optional(),
 	queueMode: z.nativeEnum(QueueMode).optional(),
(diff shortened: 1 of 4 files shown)


Victor-M-Giraldo:
I think this is in the right direction. Would just like some feedback before I continue

common/models/zod-schemas.ts
```diff
@@ -55,3 +56,44 @@ export const OttApiRequestAccountRecoveryVerifySchema = z.object({
...
+		grants: z.array(
+			z.tuple([
+				z.object({
+					role: z.nativeEnum(Role),
+				}),
+				z.object({
+					grantMask: z.number(),
+				}),
+			])
+		),
```
dyc3, line 92: This should go in `RoomSettingsSchema`

common/models/zod-schemas.ts
```diff
@@ -55,3 +56,44 @@ export const OttApiRequestAccountRecoveryVerifySchema = z.object({
...
+		...RoomSettingsSchema.shape,
+	}),
+	z.object({
+		claim: z.boolean().optional(),
+		...RoomSettingsSchema.shape,
```
dyc3, line 97: and `RoomSettingsSchema` should be excluded from here.

Victor-M-Giraldo: To clarify, did you mean they should be excluded from both? Or is how it is now fine?

Still needs to be a union type right? So union type between the claim and the rest of the room settings fields?

common/models/zod-schemas.ts
```diff
@@ -55,3 +56,40 @@ export const OttApiRequestAccountRecoveryVerifySchema = z.object({
...
+	grants: z.array(GrantSchema).optional(),
+	autoSkipSegmentCategories: z.array(CategorySchema).optional(),
+	restoreQueueBehavior: z.nativeEnum(BehaviorOption).optional(),
+	enableVoteSkip: z.boolean().optional(),
+});
+
+const ClaimSchema = z.object({
+	claim: z.boolean().optional(),
+	...RoomSettingsSchema.shape,
+});
```
dyc3, line 93: I meant just the object with `claim`. It shouldn't include room settings.
```suggestion
const ClaimSchema = z.object({
	claim: z.boolean().optional(),
});
```

common/models/zod-schemas.ts
```diff
@@ -55,3 +56,40 @@ export const OttApiRequestAccountRecoveryVerifySchema = z.object({
...
+	"preview",
+]);
+
+const GrantSchema = z.tuple([
+	z.object({
+		role: z.nativeEnum(Role),
+	}),
+	z.object({
+		grantMask: z.number(),
+	}),
```
dyc3, line 76: This doesn't look right. Its a tuple of 2 objects, not of the enum and the mask

Victor-M-Giraldo:
Just giving an update on this one. Still working on it, I'm aware I need to narrow the type down but I'm also encountering an issue after I do that. It's a WIP though.

dyc3:
IIRC the only endpoint that uses query params is `/api/data/previewAdd`, which is in `data.ts`

dyc3:
As discussed offline:

for handling `grants`, create a new object from `body`

probably something like
```
const foo = {
...body,
grants: new Grants(body.grants)
}
```

be careful to not add it if it doesn't exist on the body

Victor-M-Giraldo:
I know you said not to do `else if ('visibility' in body)` but if I just do `else` I get type errors on the grants. Also, I'm getting type errors when trying to update the settings to just be the body and it says the Grants is missing some properties.

server/api/room.ts
```diff
@@ -184,76 +184,63 @@ const getRoom: RequestHandler<{ name: string }, OttApiResponseGetRoom, unknown>
...
+	} else if ('visibility' in body) {
+		const newBody = {
+			...body,
+			grants: new Grants(body.grants),
+		}
+
 		const roomRequest: ApplySettingsRequest = {
 			type: RoomRequestType.ApplySettingsRequest,
-			settings: req.body,
+			settings: body,
```
dyc3, line 239: should be
```suggestion
			settings: newBody,
```
or
```suggestion
			settings: {
				...body,
				grants: new Grants(body.grants),
			},
```

server/api/room.ts
```diff
@@ -184,76 +184,63 @@ const getRoom: RequestHandler<{ name: string }, OttApiResponseGetRoom, unknown>
...
 		};
 
 		await room.processUnauthorizedRequest(roomRequest, { token: req.token });
+
```
dyc3, line 243: nit: stray newline

dyc3 (changes requested):

Victor-M-Giraldo:
Still need to fix tests.

dyc3 (approved):

PR #1581, opened 2024-03-29
add client ids to system state

### LINKED ISSUE #1562: vis: add client ids to system state
Currently, we just get the number of clients in each room. We want to be able to get the actual client ids for visualizing activity.

Victor-M-Giraldo:
- closes #1562

### DIFF
--- a/packages/ott-vis-panel/src/components/TreeDisplay.tsx
+++ b/packages/ott-vis-panel/src/components/TreeDisplay.tsx
@@ -77,14 +77,12 @@ function buildRoomSubtree(room: Room, region: string): TreeNode {
 		id: room.name,
 		region: region,
 		group: "room",
-		children: Array.from({ length: room.clients }, (_, index) => {
-			return {
-				id: `${room.name}-${index}`,
-				region: region,
-				group: "client",
-				children: [],
-			};
-		}),
+		children: room.clients.map(c => ({
+			id: c.id,
+			region: region,
+			group: "client",
+			children: [],
+		})),
 	};
 	return roomNode;
 }
@@ -667,6 +665,9 @@ const TreeDisplay: React.FC<TreeDisplayProps> = ({
 	const eventBus = useEventBus();
 	useEffect(() => {
 		const sub = eventBus.subscribe(event => {
+			if (event.direction !== "rx") {
+				return;
+			}
 			const node = d3.select(`[data-nodeid="${event.node_id}"]`);
 			if (node.empty()) {
 				return;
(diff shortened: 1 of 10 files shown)


Victor-M-Giraldo:
Still need to actually fix the tests.

packages/ott-vis-panel/src/components/TreeDisplay.tsx
```diff
@@ -65,7 +65,7 @@ function buildMonolithTrees(monoliths: Monolith[]): TreeNode[] {
 				id: room.name,
 				region: monolith.region,
 				group: "room",
-				children: Array.from({ length: room.clients }, (_, index) => {
+				children: Array.from({ length: room.clients.length }, (_, index) => {
 					return {
 						id: `${room.name}-${index}`,
```
dyc3, line 70: This needs to become the client id

packages/ott-vis-panel/src/components/TreeDisplay.tsx
```diff
@@ -65,7 +65,7 @@ function buildMonolithTrees(monoliths: Monolith[]): TreeNode[] {
 				id: room.name,
 				region: monolith.region,
 				group: "room",
-				children: Array.from({ length: room.clients }, (_, index) => {
+				children: Array.from({ length: room.clients.length }, (_, index) => {
```
dyc3, line 68: this `Array.from` hack can be removed in favor of a `.map` or something more readable.

dyc3 (commented):
Looking good so far.

Victor-M-Giraldo:
Don't know if the unit test is really necessary, but I think this ones good.

dyc3 (changes requested):
This is good for `ott-vis-panel`, but `ott-collector` and `ott-balancer` still need to be updated accordingly.

Currently, it seems like the interop between the collector and the panel isn't being enforced in CI.

dyc3:
You should rebase this PR for better type checking

dyc3:
Whoops, I made some changes earlier today that are conflicting. Sorry about that, I can rebase this for you if you want

Victor-M-Giraldo:
> Whoops, I made some changes earlier today that are conflicting. Sorry about that, I can rebase this for you if you want

Yeah, sure. I tried to do it myself. There is only one conflict but the incoming change is an older version of the `buildMonolithTrees` which I don't get why.

Victor-M-Giraldo:
Since `countRoomClients` is returning a map of names to number of clients, I don't think I can access the id for the RegionView and GlobalView.

dyc3 (approved):
Yeah don't worry about global/region view. We're probably going to trash them.

Victor-M-Giraldo:
Okay, but say if you did want to do that would you change `countRoomClients` to be like `Record<string, ClientState[]` or something.

dyc3:
I would probably leave that count function as is and create a new one.

PR #1760, opened 2024-04-29
add support for node 22

dyc3:


Sysadminfromhell:
@dyc3 isnt that fixed with our last update from `sodium-nativ` or do we need to update further?

dyc3:
This is probably too stale

Issue #1289, opened 2024-02-01
Bulk add previews that completely fail to fetch any videos can get their response cached


dyc3:
### Current Behavior

It's possible for a bulk add preview, one with multiple links separated by newlines, to fail to get all of the videos requested. In this case, the add preview will still succeed, and the result will get cached. This will cause the user to be unable to get a fresh add preview

### Expected Behavior

If grabbing all the videos fails, the add preview should fail with a non-200 status code, and exclude any caching headers.

### Steps To Reproduce

Recommended: use `miniserve` to serve files over http

1. In the add preview box, paste 2 links to non existent videos
2. Wait for add preview to complete
3. Bring up a webserver to serve videos to make the links in the add preview valid
4. Refresh the add preview
5. See the previous result remain

### Environment

- [X] This happens on the official site, opentogethertube.com
- [X] This happens using a self-hosted version.
- [ ] I'm using the docker image.

PR #1787, opened 2024-08-19
server: fix bulk add preview caching 

IchanP:
Fix for #1289. Also filters videos that failed to fetch from being returned to the client.

### DIFF
--- a/server/infoextractor.ts
+++ b/server/infoextractor.ts
@@ -13,6 +13,7 @@ import {
 	UnsupportedServiceException,
 	InvalidAddPreviewInputException,
 	FeatureDisabledException,
+	VideoNotFoundException,
 } from "./exceptions";
 import { getLogger } from "./logger";
 import { redisClient } from "./redisclient";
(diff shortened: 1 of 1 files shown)


server/infoextractor.ts
```diff
@@ -264,25 +265,35 @@ export default {
...
-					}
-				});
+				const finalResults = cachedVideos
+					.map(video => {
+						const fetchedVideo = fetchedVideos.find(v => v.id === video.id);
+						const videoKeys = Object.keys(video);
+						if (fetchedVideo) {
+							return mergeVideo(video, fetchedVideo);
+						} else if (videoKeys.length > 2) {
+							return video;
```
dyc3, line 280: What is this branch here for?

dyc3 (changes requested):
Thanks for the PR!

It doesn't look like this accounts for partial successes. If the add preview only partially succeeds (eg. if 2 video links are requested, but only 1 video is returned), the request will still get cached in the user's browser.

server/infoextractor.ts
IchanP: You're right it's still broken. I don't think this approach is correct, I'll revise it.

The branch is to only return the video if the video is found in the cache, since it should contain more than the VideoId fields then.

dyc3: This is incorrect behavior then. We need to be able to return videos that were not already in the cache.

Sysadminfromhell:
@dyc3 / @IchanP whats about this PR? Its pretty old and ig far behind.

dyc3:
@IchanP Still interested in this pr?

IchanP:
> @IchanP Still interested in this pr?

Sorry for the super late reply. You can write me off on this, I don't currently have the time to look into this.

dyc3:
No worries!

Issue #1828, opened 2025-04-01
Remember SponsorBlock settings


mayudev:
### Is your feature request related to a problem? Please describe your use case.

Every time a room is created the default SponsorBlock settings are applied. I think they might be a bit too aggressive, so it's good that there is an option to adjust them, but redoing that every time I create a room is a little bit frustrating. I think it'd make for a better user experience if they got remembered by the browser (or account, in case it's logged in).

### Describe the solution you'd like

The SponsorBlock settings would get remembered by the user's browser so the next time they create a room, the settings get automatically applied.

### Additional context

I can work on this.

dyc3:
That's kinda what permanent rooms are for. I'm not opposed to this necessarily if it doesn't add too much complexity.

PR #1829, opened 2025-04-03
feat(client): save sponsorblock settings in browser and apply them on room creation

mayudev:
Related to https://github.com/dyc3/opentogethertube/issues/1828

- Saves SponsorBlock settings to localStorage/settings whenever they are changed
- Applies those settings on room creation (both temporary and permanent)

### DIFF
--- a/client/src/components/RoomSettingsForm.vue
+++ b/client/src/components/RoomSettingsForm.vue
@@ -158,7 +158,6 @@ import { Ref, onMounted, reactive, ref, toRefs } from "vue";
 import { useStore } from "@/store";
 import { useI18n } from "vue-i18n";
 import { OttApiResponseGetRoom } from "ott-common/models/rest-api";
-import { ALL_SKIP_CATEGORIES } from "ott-common/constants";
 import { useGrants } from "./composables/grants";
 import { useRoute } from "vue-router";
 import { watch } from "vue";
--- a/client/src/components/RoomSettingsForm.vue
+++ b/client/src/components/RoomSettingsForm.vue
@@ -261,6 +261,13 @@ function getRoomSettingsSubmit(): Partial<RoomSettings> {
 	return _.omit(inputRoomSettings, blocked);
 }
 
+/** Save SponsorBlock settings in user's settings */
+function updateClientSettings() {
+	const copy = _.cloneDeep(store.state.settings);
+	copy.autoSkipSegmentCategories = settings.autoSkipSegmentCategories.value;
+	store.commit("settings/UPDATE", copy);
+}
+
 /** Take room settings from the UI and submit them to the server. */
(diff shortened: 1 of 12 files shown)


client/src/views/Room.vue
```diff
@@ -442,6 +443,18 @@ export default defineComponent({
...
+					await API.patch(`/room/${route.params.roomId ?? store.state.room.name}`, {
+						autoSkipSegmentCategories: store.state.settings.autoSkipSegmentCategories,
+					});
+				} catch (e) {
+					console.error(e);
+				}
+
+				delete history.state.created;
+			}
+
```
dyc3, line 457: This feels very hacky. We can send the room settings directly to the `/api/room/create` and `/api/room/generate` endpoints, and avoid the extra round trip.

client/src/stores/settings.ts
```diff
@@ -8,6 +9,7 @@ export interface SettingsState {
 	theme: Theme;
 	sfxEnabled: boolean;
 	sfxVolume: number;
+	autoSkipSegmentCategories?: Category[];
```
dyc3, line 12: For the client settings here, it would be better if we could nest this under a `defaultRoomSettings` so we can add other settings like `grants` to it.

client/src/components/RoomSettingsForm.vue
```diff
@@ -269,6 +276,7 @@ async function submitRoomSettings() {
 			`/room/${route.params.roomId ?? store.state.room.name}`,
 			getRoomSettingsSubmit()
 		);
+		updateClientSettings();
```
dyc3, line 279: There's almost certainly a race condition here. You're grabbing the settings from the store before waiting for the sync message that actually updates the store.

Regardless, I don't really like this from a UX perspective. Modifying room settings shouldn't have side effects like this. The client settings are exposed in the "Preferences" dialog, which can be found here: https://github.com/dyc3/opentogethertube/blob/master/client/src/components/ClientSettingsDialog.vue

dyc3 (changes requested):
Thanks for the PR! It's a good first draft.

client/src/components/RoomSettingsForm.vue
mayudev: That's fair... I didn't think well enough about this because it works perfectly for my use case, but it makes sense that some users might not like it.
Would you suggest adding the default SponsorBlock settings to the client settings instead?

dyc3: Yeah. We should have a "default room settings" setting in the client settings.

dyc3 (commented):
This looks great! Just a couple of minor nitpicks.

dyc3 (approved):
Great PR, thank you! Hope to see more from you!

PR #1847, opened 2025-08-15
update german language

Sebasee009:
It hasn't been updated in a long time, so I thought it was time to do so.

### DIFF
--- a/client/src/locales/de.ts
+++ b/client/src/locales/de.ts
@@ -1,18 +1,42 @@
-import { OttWebsocketError, Role } from "ott-common/models/types";
+import { OttWebsocketError, BehaviorOption, Role } from "ott-common/models/types";
 
 export default {
 	"common": {
+		"yes": "Ja",
+		"no": "Nein",
+		"ok": "OK",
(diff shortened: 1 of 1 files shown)


client/src/locales/de.ts
```diff
@@ -67,8 +91,8 @@ export default {
 			description2:
 				"Beteiligen Sie sich an der Entwicklung, indem Sie Ihre Ideen oder Ihren Code beisteuern, oder zeigen Sie Ihre Unterstützung, indem Sie Spender werden. Alle Spenden werden verwendet, um die Betriebskosten und die Entwicklung von OpenTogetherTube zu finanzieren und um OpenTogetherTube werbefrei zu halten.",
 			how: "Wie kann ich helfen?",
-			sponsor: "@:nav.support",
```
dyc3, line 70: This is intentionally a reference. Does this change make more sense contextually where it appears than the referenced text?

dyc3 (approved):
Thank you! Just a quick question

dyc3 (approved):

Issue #1842, opened 2025-07-24
Support for Invidious


Sysadminfromhell:
### Is your feature request related to a problem? Please describe your use case.

It would be greate to be able to use a Invidious instance to avoid ads.

### Describe the solution you'd like

To Support public reachable Invidious instances.

dyc3:
I remember briefly looking into this but I remember running into issues around the info extractor. Basically, OTT needs to be able to know if a given URL belongs to a Service. It normally does this by inspecting the domain name of the URL you give it. The problem is that Invidious instances could be on any domain, so OTT needs to be able to probe the given instance to see if it is really an Invidious instance. Peertube support also has this problem. Is there a simple API endpoint that OTT could ping to do this check? The result should be cached too.

**Regardless of that, if you really want to avoid ads, just install uBlock Origin.** Youtube embeds already don't run video ads, and the popup ads are easily suppressed by adblock.

Sysadminfromhell:
hey @dyc3 thanks for your reply, I have to look into that more but if you want a public instance to look for yourself you can: yt.safh.de (this is mine) feel free to use it for the dev process and beyond. I will see what I can figure out about the api

Sysadminfromhell:
Which kind of information does the api need to expose / send for ott to understand? there is the api call for stats (since alive and version and such) but nothing more afaik.
(https://docs.invidious.io/api/#get-apiv1videosid)

Sysadminfromhell:
if you provide an youtube url you can see anything about the video:

https://yt.safh.de/api/v1/videos/Wf8DxisYbYk

`{"type":"video","title":"Dieses Spiel weckt Urängste! | Sharks and Minnows","videoId":"Wf8DxisYbYk","videoThumbnails":[{"quality":"maxres","url":"https://yt.safh.de/vi/Wf8DxisYbYk/maxres.jpg","width":1280,"height":720},{"quality":"maxresdefault","url":"https://yt.safh.de/vi/Wf8DxisYbYk/maxresdefault.jpg","width":1280,"height":720},{"quality":"sddefault","url":"https://yt.safh.de/vi/Wf8DxisYbYk/sddefault.jpg","width":640,"height":480}, [...]`

dyc3:
> Which kind of information does the api need to expose?

Literally anything that indicates "Hey, I'm an Invidious instance". Ideally some kind of status endpoint that would be unique to Invidious.

[...]

PR #1849, opened 2025-08-18
First Implementation for Invidious support, tested on my instance and…

Sysadminfromhell:
First Implementation of Invidious Support,
Had to Modify mutltiple files to get the support working completly due to DASH, WEBM and HLS Support in Invidious.
Feel free to test it on your own. For me it worked fine a few times while developing this feature.

Related: #1842

### DIFF
--- a/deploy/monolith.Dockerfile
+++ b/deploy/monolith.Dockerfile
@@ -1,6 +1,6 @@
 # Optimized for layer cache hits to speed up builds
 
-FROM node:18-alpine3.19 as dep-install-stage
+FROM node:18-alpine3.19 AS dep-install-stage
 
 WORKDIR /app
 RUN corepack enable
@@ -12,7 +12,7 @@ COPY client/package.json client/
 COPY server/package.json server/
 RUN yarn workspaces focus ott-common ott-client ott-server
 
-FROM node:18-alpine3.19 as build-stage
+FROM node:18-alpine3.19 AS build-stage
 ARG GIT_COMMIT
 ENV GIT_COMMIT=$GIT_COMMIT
 
@@ -29,14 +29,14 @@ RUN yarn workspace ott-common run build && yarn workspace ott-client run build &
 RUN rm -rf packages/ott-vis*
 RUN rm -rf node_modules && yarn workspaces focus ott-server --production
 
-FROM node:18-alpine3.19 as production-stage
+FROM node:18-alpine3.19 AS production-stage
 
 WORKDIR /app
 RUN corepack enable
 COPY --from=build-stage /app /app
 RUN rm -rf client/public client/src client/.browserslistrc .eslintrc.js .gitignore client/vite.config.js client/babel.config.js docker-compose.yml /root/.npm tools crates
 
-FROM node:18-alpine3.19 as docker-stage
+FROM node:18-alpine3.19 AS docker-stage
 # For use in docker-compose
 
 WORKDIR /app
@@ -50,7 +50,7 @@ HEALTHCHECK --interval=30s --timeout=3s CMD ( curl -f http://localhost:8080/api/
 
 CMD ["/bin/sh", "wait_for_db.sh", "--", "yarn", "run", "start"]
 
-FROM node:18-alpine3.19 as deploy-stage
+FROM node:18-alpine3.19 AS deploy-stage
 # For deployment on Fly
 ARG DEPLOY_TARGET
 
--- a/deploy/monolith.Dockerfile
+++ b/deploy/monolith.Dockerfile
@@ -1,6 +1,6 @@
 # Optimized for layer cache hits to speed up builds
 
-FROM node:18-alpine3.19 AS dep-install-stage
+FROM node:20-alpine3.22 AS dep-install-stage
 
 WORKDIR /app
 RUN corepack enable
@@ -12,7 +12,7 @@ COPY client/package.json client/
 COPY server/package.json server/
 RUN yarn workspaces focus ott-common ott-client ott-server
 
-FROM node:18-alpine3.19 AS build-stage
+FROM node:20-alpine3.22 AS build-stage
 ARG GIT_COMMIT
 ENV GIT_COMMIT=$GIT_COMMIT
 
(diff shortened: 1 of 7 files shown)


Sysadminfromhell:
Build for me in 5:18 (mm:ss) just fine.

server/infoextractor.ts
```diff
@@ -27,12 +27,17 @@ import PeertubeAdapter from "./services/peertube";
...
+import InvidiousAdapter from "./services/invidious";
 
 const log = getLogger("infoextract");
 
 const ADD_PREVIEW_SEARCH_MIN_LENGTH = conf.get("add_preview.search.min_query_length");
 const ENABLE_SEARCH = conf.get("add_preview.search.enabled");
 
+const SERVICE_ALIASES: Partial<Record<VideoService, VideoService[]>> = {
+  invidious: ["hls", "dash", "direct"],
+};
```
dyc3: I don't really like the idea of having service aliases. The service name dictates which player the `OmniPlayer` component uses in order to play the video. A ServiceAdapter (on the server side) can emit videos that are attributed to any service (see for example the `tubi` service adapter, or the `direct` one even).

server/ott-config.ts
```diff
@@ -285,6 +285,35 @@ export const conf = convict({
...
+			},
+			emit_as_direct: {
+				default: true,
+				doc: "Whether to emit Invidious videos as direct or hls videos instead of Invidious videos. This is useful if the Invidious Instance embeds are broken.",
+				format: Boolean,
+				env: "INVIDIOUS_EMIT_AS_DIRECT",
+			},
+			local: {
+				default: true,
+				doc: "Use API local via Proxy",
```
dyc3: This doc comment for `local` isn't very good. It's way too brief and doesn't provide enough context.

Sysadminfromhell:
I accidently pushed 2 things here:
1. invidious support
2. Bump deploy/monolith to supported version due to:
   a. Alpine 3.19 is eol since one month already
   b. Node.JS 18 will be EOL in 1 Month
   I Bumped both and build the docker on my end w/o any errors. If needed I can push those 2 edits via different branches and then make seperate pr's.

dyc3:
I'll try to get this reviewed today

Sysadminfromhell:
@dyc3 I want to add that I didn't test if any other usability now is broken (peertube or dash in this case), I should've. Would be nice if you could test it on your own. If you want me to test it beforehand I will do so but later today / at night.

deploy/monolith.Dockerfile
dyc3, line 1: The dockerfile change requires more scrutiny. I'll do it myself in another PR.

server/infoextractor.ts
```diff
@@ -257,12 +303,14 @@ export default {
...
-						missingInfo: storage
-							.getVideoInfoFields(video.service)
-							.filter(p => !video[p]),
-					}))
+					.map(video => {
+						let missing = storage.getVideoInfoFields(video.service).filter(p => !video[p]);
+						// HLS/DASH: do not handle missing mime as missing
+						if (video.service === "hls" || video.service === "dash") {
+						missing = missing.filter(p => p !== "mime");
+						}
```
dyc3: This looks like a hack that was made to get around an intentional restriction.

dyc3 (changes requested):

dyc3:
Also, the node 20 upgrade is going to take a while. I have to go around and update all the imports.

server/ott-config.ts
Sysadminfromhell: Will make it better today.

server/infoextractor.ts
Sysadminfromhell: It's true, I had to make a workaround to allow invidious this kind of policy violation. This issue with that is that invifoius doesn't provide a working mime which makes invifoius unusable with the mime policy inplace.

deploy/monolith.Dockerfile
Sysadminfromhell: I will take it back on my end to have the node18-alpine3.19 but beware that both is eol and a security risk at some point.

Sysadminfromhell:
@dyc3 for me the upgrade worked with warnings of some frontend modules. But for server side it worked out. But yea do accordingly just wanted to test and it worked for me so I thought no work needed.

server/infoextractor.ts
Sysadminfromhell: I removed the service alias and will update the invidious.ts today so the adapter will send a direct at the end instead of hls or dash which isnt working how its implemented in the serviceadapters. After I changed that I will test it and will update the PR

server/ott-config.ts
Sysadminfromhell: I cleaned the config file and made a more clear doc about what this flag does but I think I am going to remove it completly anyways later on. I have to think about that and maybe we can talk that through how you feel about it:

```
			local: {
				default: true,
				doc: "When true, emit proxied media URLs via the configured Invidious instance (adds `local=1` / uses `/latest_version`). This serves video through the instance to avoid client-side CORS/IP issues. When false, return raw YouTube CDN URLs instead.",
				format: Boolean,
				env: "INVIDIOUS_LOCAL_API",
			}
```,

Sysadminfromhell: Yea nvmd I changed it so we will always proxy the video over invidious to not break CORS.

server/infoextractor.ts
Sysadminfromhell: I revert back all my changes to infoextracter and just add the Adapter in the adapter section ontop. This one here is removed too.

Sysadminfromhell: The issue is, I used that to get around the mime have-to policy due to invidious not sending the mime type when use HLS/DASH for better quality then 720p. I try to figure out a way to handle that in the invidious.ts instead so the mime isnt empty but something the dash/hls player can use instead.

deploy/monolith.Dockerfile
dyc3: Yup, that's why I'm working on it :)

dyc3:
Also, I thought I should mention: a design constraint that we have is that we can't stream video from OTT. Bandwidth is expensive, and I don't have the cash flow to be serving multi gb video files to multiple users.

server/infoextractor.ts
dyc3: IIRC The mime type is also a requirement of the video player. It needs to know the mime type ahead of time so it knows how to decode the video.

Sysadminfromhell: I set one in the invidious.ts now which should work and actually do. I now have to work on the player itself because the quality is really bad all the time. I will kindly ask you maybe to reject this pr and I will do a new one where all the requested changes from you is in (I am a noob with GitHub so I actually don't know how to do it correctly)

Sysadminfromhell:
> Also, I thought I should mention: a design constraint that we have is that we can't stream video from OTT. Bandwidth is expensive, and I don't have the cash flow to be serving multi gb video files to multiple users.

but OTT never streams the video directly or does it? It should play the video from the proxy in my case invidious.

Sysadminfromhell:
I just pushed my changes to my repo for that. It automatically wants to merge them here so .... Thats what I meant by "I am a github noob", therefore:
-> Inforextract.ts changes reverted
-> Invidious.ts now gives a appropriate mime-type which the pipeline and frontend player is able to understand. Also the Invidious.ts now reports supported qualities to the pipeline.
-> PlyrPlayer.vue now takes the highest possible quality for the video. This may needs work due to bitrate / traffic. Maybe it makes more sense to give the user the option to set the quality instead.

One thing I have to work on: Sometimes when working with public instances from Invidious like yewtu.be or invidious.nerdvpn.de the API is behind a DDOS Protection or rate limited. The requests are sometimes logging me out from these instances. On my instance the API is always available but I need a workaround for that somehow.

EDIT: After some digging I figured that even the Invidious Documentation states that the API needs to be publicly available thats not always the case. Therefore if we Implement it we need to documentate that the API needs to be available at any time from OTT otherwise the invidious.ts crashes. I will add some error Handling but I guess I need a real public instance to develope further.
Maybe we can dev this Fetaure but invidious self states that due to the latest changes of the YouTube API the instances are going back / shrinking. Therefore maybe we can add this for private use, like I do but it wont be used that much ig.

Sysadminfromhell:
Close this PR to create a new one

PR #1855, opened 2025-08-20
Invidious Support 

Sysadminfromhell:
Created this PR to cleanup the mess from the old PR.
Invidious.ts => Support should be stable and "good"

Infoextrator.ts => Minimal changes to add Invidious as ServiceAdapter

PlyPlayer.Vue => Minimal Changes to support better quality for DASH and HLS when available, also fixed the dash.on to official DashJS events, if not wanted I can revert this change but I like it more this way. Its just personal preference I guess, no benefit known to me.

monolith.Dockerfile => Fixed some spelling issues with newer versions of Docker while Building (AS not as and ENV=Value instead ENV VALUE)

Enhancment in question: https://github.com/dyc3/opentogethertube/issues/1842


dyc3:
With regards to how to make github PRs effectively, these are my recommendations:

- Don't commit to master and make a PR from that. Create a new branch locally, and then push that branch up to your fork to make a PR with. If you use the github cli, it's as easy as `gh pr create` while you are on your other branch.
- To update a PR, you simply `git push` new commits to your PR branch.
- Don't forget to link to any issues that the PR is related to or resolves.

Sysadminfromhell:
> With regards to how to make github PRs effectively, these are my recommendations:
> 
> * Don't commit to master and make a PR from that. Create a new branch locally, and then push that branch up to your fork to make a PR with. If you use the github cli, it's as easy as `gh pr create` while you are on your other branch.
> * To update a PR, you simply `git push` new commits to your PR branch.
> * Don't forget to link to any issues that the PR is related to or resolves.

Thank you for that information. I will do better next time.

server/infoextractor.ts
```diff
@@ -27,6 +27,7 @@ import PeertubeAdapter from "./services/peertube";
 import PlutoAdapter from "./services/pluto";
 import DashVideoAdapter from "./services/dash";
 import { Ls } from "dayjs";
+import InvidiousAdapter from "./services/invidious";
```
dyc3: With the node 20 update, I had to change the module resolution. It now requires an explicit file extention (and it can't be `.ts` for other reasons).

```suggestion
import InvidiousAdapter from "./services/invidious.js";
```

server/services/invidious.ts
```diff
@@ -0,0 +1,416 @@
+import axios from "axios";
+import { getLogger } from "../logger";
+import { ServiceAdapter } from "../serviceadapter";
+import { conf } from "../ott-config";
+import { Video, VideoMetadata, VideoService } from "ott-common/models/video";
+import { InvalidVideoIdException } from "../exceptions";
+import storage from "../storage";
```
dyc3: Same here

```suggestion
import { getLogger } from "../logger.js";
import { ServiceAdapter } from "../serviceadapter.js";
import { conf } from "../ott-config.js";
import { Video, VideoMetadata, VideoService } from "ott-common/models/video.js";
import { InvalidVideoIdException } from "../exceptions.js";
import storage from "../storage.js";
```

server/services/invidious.ts
```diff
@@ -0,0 +1,416 @@
...
+      }
+    } catch (e) {
+      log.warn(
+        `Failed to fetch/parse ${kind.toUpperCase()} manifest for logging url=${manifestUrl} err=${e instanceof Error ? e.message : e}`
+      );
+    }
+  }
+
+  /** Probe a manifest and return its top bitrate/resolution. 
+   * Needs to be there after finishing feature to give frontend all infos about quality
```
dyc3: AFAIK, the client should be able to figure out this info. I'm pretty sure Plyr gives us this info, similar to how it gives us info about captions. We shouldn't need to compute it on the server side.

server/services/invidious.ts
```diff
@@ -0,0 +1,416 @@
...
+            kind === "dash"
+              ? "application/dash+xml,text/plain;q=0.9,*/*;q=0.8"
+              : "application/vnd.apple.mpegurl,application/x-mpegURL,text/plain;q=0.9,*/*;q=0.8",
+        },
+        responseType: "text",
+        transformResponse: [(r) => r],
+      });
+      const text = typeof resp.data === "string" ? resp.data : String(resp.data);
+      if (kind === "dash") {
+        const tags = text.match(/<Representation\b[^>]*>/g) ?? [];
```
dyc3: I don't like the excessive regexing here. We already pull in `m3u8-parser` and `@liveinstantly/dash-mpd-parser` to parse these.

dyc3 (changes requested):

dyc3:
Also, I've finished the node 20 upgrade, you'll need to rebase or merge those changes

server/services/invidious.ts
Sysadminfromhell: I agree but somewhat with Invidious this information gets lost. Idk why so I had to apply this patch to let the player know which Quality to use but even then the quality of the DASHJS is bad when Seeking / scribbling most of the time. Thats why I applied also a patch to use best quality possible.

PR #1861, opened 2025-08-20
Fix index.ts error after Node20 Upgrade

Sysadminfromhell:
Fixes this error: https://github.com/dyc3/opentogethertube/issues/1860

PR #1862, opened 2025-08-20
Add Invidious Support

Sysadminfromhell:
Here is the official now finished Invidious working support in Node20.
WARNING: Can only be merged when https://github.com/dyc3/opentogethertube/pull/1861 is merged too!

Changes to the player where needed to use the higher quality, if not set like this invidious will send always a bad quality.
Maybe, if this is more clear, I could make a PlyrPlayer only for Invidious isntead.

### DIFF
--- a/client/src/components/players/PlyrPlayer.vue
+++ b/client/src/components/players/PlyrPlayer.vue
@@ -221,11 +221,56 @@ export default defineComponent({
 				};
 				videoElem.value = document.querySelector("video") as HTMLVideoElement;
 				// ...so that we can use hls.js to change the video source
-				hls = new Hls();
+				hls = new Hls({
+					capLevelToPlayerSize: false,
+					abrEwmaDefaultEstimate: 8_000_000, // ~8Mbps
+					startLevel: -1, // auto
+				});
+
 				hls.loadSource(videoUrl.value);
 				hls.attachMedia(videoElem.value);
-				hls.on(Hls.Events.MANIFEST_PARSED, () => {
+				// Helper
+				const logCurrentHlsLevel = (label: string) => {
+					try {
+						if (!hls) return;
+							const idx =
+							(hls.currentLevel ?? -1) >= 0
+								? hls.currentLevel
+								: (hls.nextLevel ?? hls.loadLevel ?? -1);
+							const lvl = idx >= 0 ? hls.levels[idx] : undefined;
+							console.info(`[hls.js] current level (${label})`, {
+							index: idx,
+							bitrateKbps: lvl?.bitrate ? Math.round(lvl.bitrate / 1000) : undefined,
+							width: lvl?.width,
+							height: lvl?.height,
+							name: (lvl as any)?.name,
+							auto: hls.autoLevelEnabled,
+							cappedToPlayerSize: hls.config?.capLevelToPlayerSize,
+							});
+					} catch {}
+				};
+
+				hls.on(Hls.Events.MANIFEST_PARSED, (_evt, data: any) => {
+					const h = hls;
+					if (!h) return; // TS: narrow Hls | undefined -> Hls
 					console.info("PlyrPlayer: hls.js manifest parsed");
+					try {
+						const levels = (data?.levels ?? h.levels ?? []).map((l: any, i: number) => ({
+						index: i,
+						bitrateKbps: l?.bitrate ? Math.round(l.bitrate / 1000) : undefined,
+						width: l?.width,
+						height: l?.height,
+						name: l?.name,
+					}));
+						console.info("[hls.js] manifest levels", levels);
+						const max = (h.levels?.length || 0) - 1;
+						if (max >= 0) {
+							h.currentLevel = max;
+							h.nextLevel = max;
+							console.info("[hls.js] forced highest level", levels[max]);
+						}
+					} catch {}
+					logCurrentHlsLevel("after MANIFEST_PARSED");
 					emit("ready");
 					captions.captionsTracks.value = playerImpl.getCaptionsTracks();
 					captions.isCaptionsEnabled.value = playerImpl.isCaptionsEnabled();
@@ -234,6 +279,36 @@ export default defineComponent({
 					console.error("PlyrPlayer: hls.js error:", event, data);
 					console.error("PlyrPlayer: hls.js inner error:", data.error);
 					emit("error");
+				});
+				hls.on(Hls.Events.LEVEL_SWITCHING, (_evt, data: any) => {
+				const cand = hls?.levels?.[data?.level];
+					console.info("[hls.js] level requested →", {
+						index: data?.level,
+						bitrateKbps: cand?.bitrate ? Math.round(cand.bitrate / 1000) : undefined,
+						width: cand?.width, height: cand?.height, name: (cand as any)?.name,
+					});
+				});
+				hls.on(Hls.Events.LEVEL_SWITCHED, (_evt, data: any) => {
+				const cur = hls?.levels?.[data?.level];
+					console.info("[hls.js] level rendered ✔", {
+						index: data?.level,
+						bitrateKbps: cur?.bitrate ? Math.round(cur.bitrate / 1000) : undefined,
+						width: cur?.width, height: cur?.height, name: (cur as any)?.name,
+					});
+				logCurrentHlsLevel("after LEVEL_SWITCHED");
+				});
+				videoElem.value.addEventListener("playing", () => {
+					logCurrentHlsLevel("HTML5 playing");
+				});
+				videoElem.value.addEventListener("seeked", () => {
+					try {
+						const max = (hls?.levels?.length || 0) - 1;
+						if (max >= 0 && hls) {
+						hls.nextLevel = max;
+						hls.currentLevel = max;
+						}
+					} catch {}
+					logCurrentHlsLevel("after seeked (forced MAX)");
 				});
 				hls.on(Hls.Events.INIT_PTS_FOUND, () => {
 					console.info("PlyrPlayer: hls.js init pts found");
@@ -258,30 +333,93 @@ export default defineComponent({
 				};
 				videoElem.value = document.querySelector("video") as HTMLVideoElement;
 				// ...so that we can use dash.js to change the video source
+				// Add EventLister for "Seeking(Seeked)"
+				videoElem.value.addEventListener("seeked", () => {
+					try {
+						const list = dash?.getBitrateInfoListFor("video") || [];
+						const maxIndex = list.length ? list.length - 1 : 0;
+						dash?.setQualityFor("video", maxIndex);
+					} catch {}
+					logCurrentDashQuality("after seeked (forced MAX)");
+				});
+				// Allow fast switching 
+				dash.updateSettings({ streaming: { fastSwitchEnabled: true } } as any);
+
+				const logCurrentDashQuality = (label: string) => {
+				try {
+					const q = dash?.getQualityFor("video");
+					const list = dash?.getBitrateInfoListFor("video") || [];
+					const it = (q != null && list[q]) ? list[q] : undefined;
+					console.info(`[dash.js] current quality (${label})`, {
+					index: q,
+					bitrateKbps: it?.bitrate ? Math.round(it.bitrate / 1000) : undefined,
+					width: it?.width,
+					height: it?.height,
+					});
+				} catch {}
+				};
+
 				dash.initialize(videoElem.value, videoUrl.value, false);
 
-				dash.on("manifestLoaded", () => {
+				// Prefer official Eventhandling from dash.js
+				dash.on(dashjs.MediaPlayer.events.MANIFEST_LOADED, () => {
 					console.info("PlyrPlayer: dash.js manifest loaded");
+					// Try highest possible quality first
+					try {
+						const list = dash?.getBitrateInfoListFor("video") || [];
+						const maxIndex = list.length ? list.length - 1 : 0;
+						dash?.setQualityFor("video", maxIndex);
+						console.info("[dash.js] forced highest quality", {
+						index: maxIndex,
+						bitrateKbps: list[maxIndex]?.bitrate ? Math.round(list[maxIndex].bitrate / 1000) : undefined,
+						width: list[maxIndex]?.width, height: list[maxIndex]?.height,
+						});
+					} catch {}
+					logCurrentDashQuality("after MANIFEST_LOADED");
 					emit("ready");
 					captions.captionsTracks.value = playerImpl.getCaptionsTracks();
 					captions.isCaptionsEnabled.value = playerImpl.isCaptionsEnabled();
 				});
-				dash.on("error", (event: unknown) => {
+				dash.on(dashjs.MediaPlayer.events.ERROR, (event: unknown) => {
 					console.error("PlyrPlayer: dash.js error:", event);
 					emit("error");
 				});
-				dash.on("playbackError", (event: unknown) => {
-					console.error("PlyrPlayer: dash.js playback error:", event);
-					emit("error");
+				dash.on(dashjs.MediaPlayer.events.PLAYBACK_PLAYING, () => {
+				logCurrentDashQuality("PLAYBACK_PLAYING");
+				console.info("PlyrPlayer: dash.js video Playing");
 				});
-				dash.on("streamInitialized", () => {
+				//*** Temporary loginfo for quality
+				dash.on(dashjs.MediaPlayer.events.QUALITY_CHANGE_REQUESTED, (e: any) => {
+					if (e?.mediaType !== "video") return;
+					const list = dash?.getBitrateInfoListFor("video") || [];
+					const cand = list[e.newQuality];
+					console.info("[dash.js] quality requested →", {
+						index: e.newQuality,
+						bitrateKbps: cand?.bitrate ? Math.round(cand.bitrate / 1000) : undefined,
+						width: cand?.width, height: cand?.height,
+						reason: e?.reason,
+					});
+				});
+				dash.on(dashjs.MediaPlayer.events.QUALITY_CHANGE_RENDERED, (e: any) => {
+				if (e?.mediaType !== "video") return;
+					const list = dash?.getBitrateInfoListFor("video") || [];
+					const cand = list[e.newQuality];
+					console.info("[dash.js] quality rendered ✔", {
+						index: e.newQuality,
+						bitrateKbps: cand?.bitrate ? Math.round(cand.bitrate / 1000) : undefined,
+						width: cand?.width, height: cand?.height,
+					});
+				logCurrentDashQuality("after QUALITY_CHANGE_RENDERED");
+				});
+				// TEMP END ***
+				dash.on(dashjs.MediaPlayer.events.STREAM_INITIALIZED, () => {
 					console.info("PlyrPlayer: dash.js stream initialized");
 				});
-				dash.on("bufferStalled", () => {
+				dash.on(dashjs.MediaPlayer.events.BUFFER_EMPTY, () => {
 					console.info("PlyrPlayer: dash.js buffer stalled");
 					emit("buffering");
 				});
-				dash.on("bufferLoaded", () => {
+				dash.on(dashjs.MediaPlayer.events.BUFFER_LOADED, () => {
 					console.info("PlyrPlayer: dash.js buffer loaded");
 					emit("ready");
 				});
--- a/server/ott-config.ts
+++ b/server/ott-config.ts
@@ -285,6 +285,17 @@ export const conf = convict({
 				env: "PEERTUBE_EMIT_AS_DIRECT",
 			},
 		},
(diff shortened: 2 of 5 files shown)


dyc3:
Please run `yarn lint` and resolve the issues to get the CI to pass (specifically the `lint` job, theres a permissions issue with the bundle size job unfortunately)

Also FYI, github has specific syntax to make it so that when you merge a PR, it closes the linked issue: https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

server/services/invidious.ts
```diff
@@ -0,0 +1,311 @@
...
+  canHandleURL(link: string): boolean {
+    const url = new URL(link);
+    if (!this.allowedHosts.includes(url.host)) return false;
+
+    // /watch?v=VIDEOID  OR  /w/VIDEOID
+    if (url.pathname === "/watch") {
+      return url.searchParams.has("v");
+    }
+    return /^\/w\/[A-Za-z0-9_-]{11}$/.test(url.pathname);
+  }
```
dyc3: I would like to see unit tests for the URL parsing. It's easy and it provides context for what kind of links are supposed to be accepted.

server/services/invidious.ts
```diff
@@ -0,0 +1,311 @@
...
+      id = url.searchParams.get("v");
+    } else if (/^\/w\/[A-Za-z0-9_-]{11}$/.test(url.pathname)) {
+      id = url.pathname.split("/").pop() || null;
+    }
+
+    if (!id) {
+      throw new InvalidVideoIdException(this.serviceId, link);
+    }
+    return `${url.host}:${id.trim()}`;
+  }
```
dyc3: Same here, unit tests for parsing

server/services/invidious.ts
```diff
@@ -0,0 +1,311 @@
...
+
+  canHandleURL(link: string): boolean {
+    const url = new URL(link);
+    if (!this.allowedHosts.includes(url.host)) return false;
+
+    // /watch?v=VIDEOID  OR  /w/VIDEOID
+    if (url.pathname === "/watch") {
+      return url.searchParams.has("v");
+    }
+    return /^\/w\/[A-Za-z0-9_-]{11}$/.test(url.pathname);
```
dyc3: Using regexes like this causes the regex object to be recreated every time the function is run. Move this regex to a top level `const` definition. As in, in the top level scope of this file, you define a `const FOO_REGEX = ...` and then reuse that const in the functions.

Also, we don't really want to limit the id to 11 chars. If youtube decides to add another char to the ID, it would make all those videos invalid. There's no spec for youtube IDs so we have to be a little loose with our validation in favor of being a little more resilient.

server/services/invidious.ts
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    // Prefer DASH/HLS (higher quality) and fall back to progressive "direct"
+    const video = await this.parseAsDirect(data, host, id);
+
+    // Pre-fill cache for the emitted (direct) video to avoid a follow-up metadata fetch
+    // by DirectVideoAdapter (which might fail on CORS/HEAD/mime detection).
+    try {
+      await storage.updateVideoInfo(video);
+    } catch (e: any) {
```
dyc3: Prefer `unknown` over `any`

server/services/invidious.ts
```diff
@@ -0,0 +1,311 @@
...
+      const [dashProbe, hlsProbe] = await Promise.all([
+        this.probeManifest(host, id, "dash"),
+        this.probeManifest(host, id, "hls"),
+      ]);
+      const pick =
+        dashProbe && hlsProbe
+          ? (dashProbe.topKbps >= hlsProbe.topKbps ? dashProbe : hlsProbe)
+          : (dashProbe ?? hlsProbe);
+      if (pick) {
+        log.info("Picked streaming manifest", { kind: pick.kind, url: pick.url, topKbps: pick.topKbps, topRes: pick.topRes });
```
dyc3:
```suggestion
        log.debug("Picked streaming manifest", { kind: pick.kind, url: pick.url, topKbps: pick.topKbps, topRes: pick.topRes });
```

server/services/invidious.ts
```diff
@@ -0,0 +1,311 @@
...
+      let mime =
+        chosen.type?.split(";")[0]?.trim() ||
+        (chosen.container === "webm" ? "video/webm"
+          : chosen.container === "mp4" ? "video/mp4"
+            : undefined);
+      // Always proxy via the instance (avoids CORS/signature churn) and carry itag if known.
+      const proxied = this.proxiedProgressive(host, id, chosen.itag);
+      // If MIME couldn’t be inferred, default to MP4 (works best across players).
+      if (!mime) {
+        if (typeof chosen.itag === "number" && [22, 18, 59].includes(chosen.itag)) {
```
dyc3: `[22, 18, 59]` are magic numbers. Add some comments to explain them, they need a bit more documentation.

server/services/invidious.ts
```diff
@@ -0,0 +1,311 @@
...
+    if (list.length) {
+      const mp4 = list.filter(
+        f => f.container === "mp4" || (f.type || "").toLowerCase().includes("mp4")
+      );
+      // Sort by quality descending (e.g., 720 > 480 > 360)
+      const byQualityDesc = (a: InvidiousFormat, b: InvidiousFormat) =>
+        q(b.qualityLabel) - q(a.qualityLabel);
+
+      const bestMp4 = mp4.sort(byQualityDesc)[0];
+      const bestAny = list.sort(byQualityDesc)[0];
```
dyc3: Would be better to do [`_.maxBy`](https://lodash.com/docs/#maxBy). Sorting requires more operations on average.

server/services/invidious.ts
```diff
@@ -0,0 +1,311 @@
...
+            kind === "dash"
+              ? "application/dash+xml,text/plain;q=0.9,*/*;q=0.8"
+              : "application/vnd.apple.mpegurl,application/x-mpegURL,text/plain;q=0.9,*/*;q=0.8",
+        },
+        responseType: "text",
+        transformResponse: [(r) => r],
+      });
+      const text = typeof resp.data === "string" ? resp.data : String(resp.data);
+      if (kind === "dash") {
+        const tags = text.match(/<Representation\b[^>]*>/g) ?? [];
```
dyc3: Again, we already pull in libraries to parse both hls `m3u8` streams, and dash `mpd` streams. Don't regex them, it's unreliable. They must be parsed properly.

client/src/components/players/PlyrPlayer.vue
```diff
@@ -221,11 +221,56 @@ export default defineComponent({
 				};
 				videoElem.value = document.querySelector("video") as HTMLVideoElement;
 				// ...so that we can use hls.js to change the video source
-				hls = new Hls();
+				hls = new Hls({
+					capLevelToPlayerSize: false,
+					abrEwmaDefaultEstimate: 8_000_000, // ~8Mbps
+					startLevel: -1, // auto
+				});
```
dyc3: Why are we changing these options?

dyc3 (changes requested):
Thanks for working on this.

Overall, there needs to be more documentation around this feature. There's a lot of weird things going on, and it's not clear if something is an invidious quirk, a youtube quirk, a browser thing, etc. Please keep in mind that I have _absolutely no context_ for how Invidious works, and I'm relying on you to give me that context.

And please fix the lints :)

client/src/components/players/PlyrPlayer.vue
Sysadminfromhell: This was to ensure the player uses also with hls the highest possible quality. I can revert this change. I thought "If I do it on DASH for Invidious we could do it for HLS too in the same time". Just a QoS Change if you want to say.

server/services/invidious.ts
Sysadminfromhell: I changed the regex in total to this:

const INVIDIOUS_SHORT_WATCH_RE = /^\/w\/[A-Za-z0-9_-]+$/;

return INVIDIOUS_SHORT_WATCH_RE.test(url.pathname);

Sysadminfromhell:
Due to my change to use an xmlparse, which worked fine locally but not in the docker I had to add the file via yarn workspace ott-server add @xmldom/xmldom, this is forbidden in your repo which is fine. We can discuss that.

client/src/components/players/PlyrPlayer.vue
dyc3: Judging by [this](https://github.com/video-dev/hls.js/issues/5949), I don't really feel comfortable setting `abrEwmaDefaultEstimate` without doing some work to actually give it a good value, which will be different for every user. Every user has different internet speeds.

Also, defaulting to the highest quality (and not limiting it to player size) doesn't feel like we'd be acting as good stewards of the open web. Bandwidth costs money.

Lets revert this for now, and whatever equivalent there is for DASH.

client/src/components/players/PlyrPlayer.vue
```diff
@@ -13,6 +13,22 @@ import "plyr/src/sass/plyr.scss";
...
+};
+type HlsManifestParsedData = { levels?: HlsLevelLike[] };
+type HlsLevelSwitchData = { level: number };
+
+type DashQualityChangeEvent = {
+	mediaType?: string;
+	newQuality: number;
+	reason?: unknown;
+};
+
```
dyc3: Do the hls and dash packages have `@types/...` type definition packages? that would be ideal

server/services/invidious.ts
```diff
@@ -0,0 +1,382 @@
...
+				},
+				responseType: "text",
+				transformResponse: [r => r],
+			});
+			const text = typeof resp.data === "string" ? resp.data : String(resp.data);
+			if (kind === "dash") {
+				// Parse MPD via xmldom (server-side safe; no browser DOM in Node). – Environment constraint
+				type XmlEl = { getAttribute(name: string): string | null };
+				const doc = new DOMParser().parseFromString(text, "application/xml");
+				const reps: XmlEl[] = Array.from(doc.getElementsByTagName("Representation") as any);
```
dyc3: This is better but still -- we already pull in `@liveinstantly/dash-mpd-parser` which should give you better objects to work with. Why can't we use that package?

yarn.lock
```diff
@@ -6813,6 +6813,13 @@ __metadata:
...
+"@xmldom/xmldom@npm:^0.8.11":
+  version: 0.8.11
+  resolution: "@xmldom/xmldom@npm:0.8.11"
+  checksum: 10c0/e768623de72c95d3dae6b5da8e33dda0d81665047811b5498d23a328d45b13feb5536fe921d0308b96a4a8dd8addf80b1f6ef466508051c0b581e63e0dc74ed5
+  languageName: node
+  linkType: hard
+
 "@xmldom/xmldom@npm:^0.8.3":
   version: 0.8.6
   resolution: "@xmldom/xmldom@npm:0.8.6"
```
dyc3: ... as you can see here in the lockfile, we already pull in this package via `@liveinstantly/dash-mpd-parser`, so this just means we have more packages and `node_modules` gets bigger.

server/services/invidious.ts
```diff
@@ -0,0 +1,382 @@
...
+
+			// Known YouTube itag→MIME hints for progressive formats (historical values; may vary over time). – YouTube-specific
+			const ITAG_TO_MIME: Record<number, string> = {
+				18: "video/mp4", // 360p
+				22: "video/mp4", // 720p
+				59: "video/mp4", // 480p
+				43: "video/webm", // 360p (VP8/Vorbis)
+				44: "video/webm", // 480p
+				45: "video/webm", // 720p
+			};
```
dyc3: This is a constant it should go at the top level. This should also be a `Map` instead.

dyc3 (changes requested):

client/src/components/players/PlyrPlayer.vue
Sysadminfromhell: @dyc3 as you say boss. I reverted all my changes to the file but I want you to check the Quality afterwards with my public invidious instance. The quality in OTT must be 480p or less. Because there is no control I assume that the PlyrPlayer.vue takes fifo attempt and the first from top to bot is the lowest one. It would be more "nice" if the controlpanel would've a controlsetting for quality.

client/src/components/players/PlyrPlayer.vue
Sysadminfromhell: I reverted all my changes to the PlyrPlayer.vue but have a copy jic. If we going to work on the quality setting in the future we can use that and put it into types.

yarn.lock
Sysadminfromhell: For some reason the import didnt work so I assumed its not there. Yarn told me to install it on my local maschine idk why the Import didnt work so I didnt check if its in the yarn.lock / package.json.

Sysadminfromhell: Fixed that issue

server/services/invidious.ts
Sysadminfromhell: Done

server/services/invidious.ts
Sysadminfromhell: The issue with that one was until now this was json and the xmlparser was, xml. I had to rework the whole code to get it to work with the old 0.5.0 MPDParser but it shoud work now. Yarn Lint und Yarn Test ist done. I had to modify the test file too of course. Will be pushed once I build the docker and tested it.

Sysadminfromhell:
@dyc3 anything you need from me further?

dyc3:
@Sysadminfromhell I'll review this again this weekend

server/ott-config.ts
```diff
@@ -285,6 +285,17 @@ export const conf = convict({
 				env: "PEERTUBE_EMIT_AS_DIRECT",
 			},
 		},
+		invidious: {
+			instances: {
+				default: ["yt.safh.de"],
```
dyc3: Can we add a couple more default instances? Because this currently affects what links are accepted.

Good followup work for this PR would be to be able to probe invidious/peertube instances, and cache the result so that we can have the main site not rely on explicit configuration, and people could supply their own instance to stream from.

server/services/invidious.ts
```diff
@@ -0,0 +1,426 @@
...
+
+// Known YouTube itag→MIME hints for progressive formats (historical values; may vary over time). – YouTube-specific
+const ITAG_TO_MIME = new Map<number, string>([
+	[18, "video/mp4"], // 360p
+	[22, "video/mp4"], // 720p
+	[59, "video/mp4"], // 480p
+	[43, "video/webm"], // 360p (VP8/Vorbis)
+	[44, "video/webm"], // 480p
+	[45, "video/webm"], // 720p
+]);
```
dyc3: Are these documented anywhere? It would be good to be able to link to some resource in a comment.

server/services/invidious.ts
```diff
@@ -0,0 +1,426 @@
...
+		if (!thumbnails || thumbnails.length === 0) {
+			return undefined;
+		}
+		// YouTube commonly exposes multiple thumbnail sizes; prefer the largest. – YouTube behavior
+		const sorted = [...thumbnails].sort((a, b) => {
+			const aPixels = (a.width || 0) * (a.height || 0);
+			const bPixels = (b.width || 0) * (b.height || 0);
+			return bPixels - aPixels;
+		});
+		return sorted[0]?.url;
```
dyc3: This is a sort then pick max. Better to do a `_.maxBy` for less comparisons.

server/services/invidious.ts
```diff
@@ -0,0 +1,426 @@
...
+			thumbnail: this.pickBestThumbnail(inv.videoThumbnails),
+		};
+
+		// 1) Probe DASH & HLS and pick the manifest whose top variant has the higher bitrate.
+		try {
+			// Both may exist or be omitted depending on upstream availability and instance config. – YouTube/Invidious variability
+			const [dashProbe, hlsProbe] = await Promise.all([
+				this.probeManifest(host, id, "dash"),
+				this.probeManifest(host, id, "hls"),
+			]);
```
dyc3: For this case, if one of the probe fails, then the entire operation fails. This is because `Promise.all` fails if any of the promises it's handling fails (see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all)

Your comment suggests that either of these could fail independently, hence my concern. We only want to outright fail if we can do none of the probes. I think you might want `Promise.any` instead

server/services/invidious.ts
```diff
@@ -0,0 +1,426 @@
...
+			const text = typeof resp.data === "string" ? resp.data : String(resp.data);
+			if (kind === "dash") {
+				// 1) Faster, namespace- scan <Representation .../>-attr
+				let topBw = 0;
+				let topRes: string | undefined;
+
+				const repTagRe = /<Representation\b[^>]*\/?>/gi;
+				const attrNum = (src: string, name: string): number => {
+					// exakter Attributname (verhindert 'width' in 'bandwidth')
+					const re = new RegExp(`(?:^|\\s)${name}(?=\\s*=)\\s*=\\s*["'](\\d+)["']`, "i");
```
dyc3: It appears we've regressed here regarding regex instead of real parsing. I especially don't like the dynamic regex.

1. We shouldn't be building dynamic regexs
2. I realize I may have misled you a little [here](https://github.com/dyc3/opentogethertube/pull/1862#discussion_r2292278733). What I meant was that we already pull in the package, so we can just add it to our direct dependencies and be able to start using it. And we can just use the slightly outdated version without bloating `node_modules` (as long as there's no security advisories, of course).

dyc3 (changes requested):
I tested it on my end and it works fine. Just quality concerns here.

server/services/invidious.ts
Sysadminfromhell: Added more documentation to that.

server/ott-config.ts
Sysadminfromhell: I try my best to implement this but as mentioned before, the most public instances have their API behind a wall or authentication.

Sysadminfromhell: Implemented and tested on my intance.

server/ott-config.ts
```diff
@@ -285,6 +285,37 @@ export const conf = convict({
...
+					doc: "If true, accept unknown hosts and attempt to probe them as Invidious instances.",
+					format: Boolean,
+					env: "INVIDIOUS_AUTODISCOVER",
+				},
+				cache_ttl_ms: {
+					default: 300000, // 5 minutes
+					doc: "TTL for per-host probe cache in milliseconds.",
+					format: Number,
+					env: "INVIDIOUS_AUTODISCOVER_TTL_MS",
+				},
```
dyc3: I would have preferred this to be in a **follow up** PR. Putting it in this one makes this take longer to review and merge.

Sysadminfromhell: omw to rm this commit/patch

Sysadminfromhell: @dyc3 you want to open an issue for the autodiscover support or should I?

dyc3: If you think it needs more discussion first, go ahead and open an issue.

server/services/invidious.ts
```diff
@@ -0,0 +1,472 @@
...
+				headers: { Accept: "application/json" },
+			});
+			data = resLocal.data;
+		} catch {
+			// Some instances gate `local=1` behind anti-DDoS/rate-limit checks; fallback to non-proxied metadata. – Invidious infra quirk
+			const res = await this.api.get<InvidiousApiVideo>(baseUrl, {
+				headers: { Accept: "application/json" },
+			});
+			data = res.data;
+		}
```
dyc3: I tried one of the other instances (yewtu.be) and immediately got a 429 for these probes. This results in OTT returning a 500, which presents a rather unhelpful unknown error to the user. I think it would be good to at least be able to tell the user that the upstream rate limited us.

Sysadminfromhell: I was experiencing this error also before, I will work on that today.

Sysadminfromhell: I will extend the exceptin.ts accordingly. I fixed it in the Log but the user still has this weird error "Failed to get add preview. This is probably a bug, check console for details" therefore I will extend the file accordingly.

server/services/invidious.ts
```diff
@@ -163,21 +163,68 @@ export default class InvidiousAdapter extends ServiceAdapter {
...
+						throw new UpstreamInvidiousException({
+							host,
+							status: e2.response.status,
+							endpoint: baseUrl,
+						});
+					}
+					// Network / timeout (no response): surface a user-friendly 502/504
+					const code = String(e2.code || "").toUpperCase();
+					const isTimeout =
+						code === "ECONNABORTED" || /\btimeout\b/i.test(e2.message || "");
```
dyc3: nit: We don't need to regex the message. checking the code is fine. Prefer simplicity over being hyper defensive.

dyc3: nit: can we simplify this whole thing to just check `e2.code` without transforming it at all?

PR #1863, opened 2025-08-31
Server: Add AutoDiscover Support for Invidious

Sysadminfromhell:
This PR reintroduces auto-discover for Invidious instances. It probes the instance’s public API, which must be reachable for this to work. If an endpoint is reachable over HTTPS but cannot be used in OTT, it’s because the instance’s API is not publicly exposed/accessible to OTT.

All changes are linted & tested:

- [x] yarn lint 
- [x] yarn test (new & existing tests pass)

`ott-config.ts` was extended to support this change.
(env file example)

```
[info_extractor.invidious]
instances=["yt.safh.de", "inv.nadeko.net", "yewtu.be"]

[info_extractor.invidious.auto_discover]
enabled=true                 # default: false
cache_ttl_ms=300000          # default: 300000 (5 min)
```

### DIFF
--- a/server/tests/unit/services/invidious.spec.ts
+++ b/server/tests/unit/services/invidious.spec.ts
@@ -2,6 +2,14 @@ import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
 import InvidiousAdapter, { INVIDIOUS_SHORT_WATCH_RE } from "../../../services/invidious.js";
 import { InvalidVideoIdException, UpstreamInvidiousException } from "../../../exceptions.js";
 
+// Hoist a single mock function so it's available inside the mock factory.
+const { lookupMock } = vi.hoisted(() => ({ lookupMock: vi.fn() }));
+vi.mock("node:dns/promises", () => ({
+	default: { lookup: lookupMock },
+	lookup: lookupMock,
+}));
+import * as dns from "node:dns/promises";
+
 vi.mock("../../../../server/redisclient.js", () => ({
 	redisClient: {
 		get: vi.fn(), // configured per test
@@ -13,11 +21,21 @@ vi.mock("../../../../server/redisclient.js", () => ({
 import { redisClient } from "../../../../server/redisclient.js";
 const redisMock: any = redisClient as any;
 
+// Helper: shape dns.lookup result depending on "all" option.
+function setLookup(address: string) {
+	const rec = { address, family: address.includes(":") ? 6 : 4 };
+	// Ensure both default.lookup and named lookup use the same implementation
+	(dns as any).lookup.mockImplementation?.((_host: string, opts?: { all?: boolean }) =>
+		Promise.resolve(opts && opts.all ? [rec] : rec)
+	);
+}
+
 describe("InvidiousAdapter (unit)", () => {
 	afterEach(() => {
 		redisMock.get?.mockReset?.();
 		redisMock.setEx?.mockReset?.();
 		redisMock.set?.mockReset?.();
+		(dns as any).lookup?.mockReset?.();
 		vi.clearAllMocks();
 		vi.restoreAllMocks();
 	});
@@ -60,33 +78,36 @@ describe("InvidiousAdapter (unit)", () => {
 			ad.allowedHosts = ["inv.nadeko.net"];
 			ad.deniedHosts = [];
 
-			expect(ad.canHandleURL("https://unknown.example/watch?v=abc123")).toBe(false);
-			expect(ad.canHandleURL("https://unknown.example/w/abc123")).toBe(false);
+			expect(ad.canHandleURL("https://unknown.example.com/watch?v=abc123")).toBe(false);
+			expect(ad.canHandleURL("https://unknown.example.com/w/abc123")).toBe(false);
 		});
 
 		it("probeForPresence accepts unknown hosts when auto-discover is enabled & instance looks healthy", async () => {
 			// Redis miss so we actually hit the network
 			redisMock.get.mockResolvedValueOnce(null);
+			// DNS resolves to a public IP (required by SSRF hardening)
+			setLookup("93.184.216.34");
 			vi.spyOn(adapter.api, "get").mockResolvedValueOnce({ status: 200, data: {} } as any);
 			const ok = await adapter.probeForPresence(
-				new URL("https://unknown.example/watch?v=abc123")
+				new URL("https://unknown.example.com/watch?v=abc123")
 			);
 			expect(ok).toBe(true);
 		});
 
 		it("probeForPresence rejects malformed URLs (missing v param)", async () => {
-			const ok = await adapter.probeForPresence(new URL("https://unknown.example/watch"));
+			const ok = await adapter.probeForPresence(new URL("https://unknown.example.com/watch"));
 			expect(ok).toBe(false);
 		});
 
 		it("probeForPresence caches results for TTL and avoids repeated network calls", async () => {
 			// 1) Cache miss: Redis GET -> null; API call -> 200; Redis SETEX called
 			redisMock.get.mockResolvedValueOnce(null);
+			setLookup("93.184.216.34");
 			const apiGet = vi
 				.spyOn(adapter.api, "get")
 				.mockResolvedValue({ status: 200, data: {} } as any);
 
-			const url = new URL("https://unknown.example/watch?v=abc123");
+			const url = new URL("https://unknown.example.com/watch?v=abc123");
 			const first = await adapter.probeForPresence(url);
 			expect(first).toBe(true);
 			expect(apiGet).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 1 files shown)


server/ott-config.ts
```diff
@@ -295,6 +295,21 @@ export const conf = convict({
...
+			// Auto-discover support: optionally accept unknown hosts and probe them.
+			auto_discover: {
+				enabled: {
+					default: false,
+					doc: "Allow tentatively accepting unknown Invidious hosts and probing them. If false, only configured instances are allowed.",
+					format: Boolean,
+					env: "INVIDIOUS_AUTO_DISCOVER_ENABLED",
+				},
+				cache_ttl_ms: {
+					default: 300000, // 5 minutes
```
dyc3: This is too short. The minimum ttl should be an hour at least. This is because we don't want to let OTT become a vector for HTTP amplification attacks.

server/ott-config.ts
```diff
@@ -295,6 +295,21 @@ export const conf = convict({
 					format: String,
 				},
 			},
+			// Auto-discover support: optionally accept unknown hosts and probe them.
+			auto_discover: {
```
dyc3, line 346: There should also be a deny list so that users can tell OTT to never probe certain domains.

server/services/invidious.ts
```diff
@@ -86,7 +97,26 @@ export default class InvidiousAdapter extends ServiceAdapter {
...
 	async initialize(): Promise<void> {
 		this.allowedHosts = conf.get("info_extractor.invidious.instances");
+		// Auto-discover config is optional; guard reads so missing keys don’t throw.
+		try {
+			this.autoDiscoverEnabled = !!conf.get("info_extractor.invidious.auto_discover.enabled");
+		} catch {
+			/* keep default false */
+		}
+		try {
+			const ttl = Number(conf.get("info_extractor.invidious.auto_discover.cache_ttl_ms"));
```
dyc3: nit: I'm pretty sure it's already a number. No parsing needed.

```suggestion
			const ttl = conf.get("info_extractor.invidious.auto_discover.cache_ttl_ms");
```

server/services/invidious.ts
```diff
@@ -103,9 +133,13 @@ export default class InvidiousAdapter extends ServiceAdapter {
...
 		const hostAllowed =
 			this.allowedHosts.includes(url.host) || this.allowedHosts.includes(url.hostname);
-		if (!hostAllowed) {
+		if (!this.autoDiscoverEnabled && !hostAllowed) {
 			return false;
 		}
+		if (this.autoDiscoverEnabled && !hostAllowed) {
+			// Let it through tentatively; we’ll probe in the background during parsing.
+			log.debug(`invidious:autodiscover tentative accept host=${url.host}`);
+		}
```
dyc3: This approach won't work. It makes the probing dependent on the order in which `ServiceAdapter`s are evaluated, and in the worst case, blocks subsequent adapters from ever running. Ideally, we extend the `ServiceAdapter` interface in some way to accommodate this new feature. Peertube has a similar use case.

server/services/invidious.ts
```diff
@@ -76,6 +76,17 @@ export default class InvidiousAdapter extends ServiceAdapter {
...
 
+	/** When true, accept links from hosts not listed in config and probe them. */
+	private autoDiscoverEnabled = false;
+	/** Probe result cache TTL (ms). */
+	private autoDiscoverTTL = 300000; // 5 minutes
+	/**
+	 * Simple in-memory probe cache: host → { ok, ts }.
+	 * ok=true means the host recently answered a known Invidious endpoint.
+	 */
+	private hostProbeCache = new Map<string, { ok: boolean; ts: number }>();
```
dyc3: We can't have this be an in memory cache. It must be in Redis. OTT can be scaled horizontally, which means multiple instances of the monolith can be run side by side and act as effectively one web app, and this is how the official deployment works.

It will also make the logic a lot more simple because Redis can handle the expiration and key eviction.

dyc3 (changes requested):

server/services/invidious.ts
Sysadminfromhell: Here’s a lean, backward-compatible approach with a second (tentative) match stage that avoids adapter-ordering pitfalls: we add optional hooks to ServiceAdapter and adjust the infoextractor. The Invidious adapter stays strict in canHandleURL and uses the new tentative stage to implement auto-discover. This would make the patch bigger but future proof. What do you say?

server/ott-config.ts
Sysadminfromhell: I set default to 1 hour and this is the minimum that can be set. So not shorter then 1 hour then.

server/services/invidious.ts
```diff
@@ -7,6 +7,7 @@ import { DashMPD } from "@liveinstantly/dash-mpd-parser";
 import { conf } from "../ott-config.js";
 import { Video, VideoMetadata, VideoService } from "ott-common/models/video.js";
 import { InvalidVideoIdException, UpstreamInvidiousException } from "../exceptions.js";
+import redis from "redis";
```
dyc3: You need to import from [`redisclient`](https://github.com/dyc3/opentogethertube/blob/master/server/redisclient.ts) (this will also fix the type issues)

server/services/invidious.ts
dyc3: Don't take this the wrong way, but is this an AI response?

Regardless, I don't understand what you mean. I am saying that probing does need to be generic, and not specific to invidious.

Sysadminfromhell: It's not ai it's translated. I sometimes need to translate what you mean to my language and vise-versa. I try not to use it if it's to confusing in the future.

dyc3: Not a problem. For the record, I'm not opposed to using AI tools, as long as it's disclosed appropriately. I don't want to just be talking to some AI through a middle man. :)

server/services/invidious.ts
Sysadminfromhell: Ah okay, I was copying this from the youtube.ts and the test worked. I will have a deeper look into that because its my first time working with redis.

server/services/invidious.ts
Sysadminfromhell: What I meant yesterday was that we could extend the ServiceAdapter logic with a CanMayHandle function with true and false and if the serviceAdapter suppors it the function is true. if true we could use an autodiscover mechanism for these ServiceAdapters who support it in the future.

Sysadminfromhell: In 2nd thought I dont know how to implement that w/o blocking the pipe as you mentioned earlier. I need to analyze and have a look into the ServiceAdapter logic.

Sysadminfromhell: Something like this:

ServiceAdapter.ts

```
45	/**
46	 * 
47	 * Returns true if this service adapter supports Autodiscover on given link.
48	 */
49	canAutodiscover(link: string): boolean {
50		return false;
51	}
```
Invidious.ts

```
128	canAutodiscover(link: string): boolean {
129		if(this.autoDiscoverEnabled) {
130			return true;
131		} else {
132			return false;
133		}
134	}
```
Needs work but more/less this logic maybe.

Sysadminfromhell: further I would extend the serviceadapter with the autodiscover function so its not exclusive to invidious and generic. Therefore we could implement that for other services too if needed.

Sysadminfromhell: I will push these changes so you can review them because I am really not that sure if this logic would work like this. I feel like I am missing a piece somewhere (same feeling I had with the excpetions).

Sysadminfromhell:
I am Thinking to close this PR and split it:
1. Generic Autodiscover function
2. Invidious Autodiscover function

Or you think I could do it in one big PR then? I follow your preference here.

dyc3:
Whatever is easier for you.

server/ott-config.ts
```diff
@@ -295,6 +295,47 @@ export const conf = convict({
...
+					// Enforce a lower bound of 1 hour even if the env/config provides a smaller value.
+					// `coerce` runs before format validation in convict, so we clamp here.
+					coerce: (v: unknown) => {
+						const ONE_HOUR = 3600000;
+						const n = typeof v === "number" ? v : Number(v);
+						if (!Number.isFinite(n)) {
+							return ONE_HOUR;
+						}
+						return n < ONE_HOUR ? ONE_HOUR : n;
+					},
```
dyc3: We shouldn't coerce the value into the valid range. Silently not doing what the user wants is bad UX. We need to emit a validation error.

server/ott-config.ts
```diff
@@ -295,6 +295,47 @@ export const conf = convict({
...
+				format: Array,
+				env: "INVIDIOUS_DENY_LIST",
+				children: { format: String },
+				// Normalize to lowercase & trim, drop empty values.
+				coerce: (v: unknown) => {
+					if (!Array.isArray(v)) {
+						return [];
+					}
+					return v.map(s => String(s).toLowerCase().trim()).filter(Boolean);
+				},
```
dyc3: Coercing to lowercase makes sense here, but configuration should still emit validation errors for invalid entries instead of filtering them out.

server/serviceadapter.ts
```diff
@@ -42,6 +42,13 @@ export class ServiceAdapter {
 		return false;
 	}
 
+	/**
+	 * Returns true if this service adapter supports Autodiscover on given link.
+	 */
+	canAutodiscover(link: string): boolean {
+		return false;
+	}
+
```
dyc3: I think this would be better:

1. `ServiceAdapter` base class defines the probe interface:
```
get canProbePresence(): boolean
async probeForPresence(url: Url): boolean
```
2. Info extractor handles the core logic (starting the probe, caching the result), and starts probing if the link after it fails to match the link against any of the adapters.

server/services/invidious.ts
```diff
@@ -76,6 +77,13 @@ export default class InvidiousAdapter extends ServiceAdapter {
 
 	allowedHosts: string[] = [];
 
+	private deniedHosts: string[] = [];
+
+	/** When true, accept links from hosts not listed in config and probe them. */
+	private autoDiscoverEnabled = false;
+	/** Probe result cache TTL (ms). */
+	private autoDiscoverTTL = 3600000; // time in ms = 1 hour
+
```
dyc3: We should just always read these from the config.

server/services/invidious.ts
```diff
@@ -86,7 +94,42 @@ export default class InvidiousAdapter extends ServiceAdapter {
 
 	async initialize(): Promise<void> {
 		this.allowedHosts = conf.get("info_extractor.invidious.instances");
+		// Load deny-list (optional); keep normalized to lowercase
+		try {
+			this.deniedHosts = (conf.get("info_extractor.invidious.deny_list") as string[]).map(h =>
+				String(h).toLowerCase()
```
dyc3: the values are already strings, and lowercased by the time we get here (because of the config validation/coersion)

Sysadminfromhell:
> Whatever is easier for you.

I didnt expect you to answer this quick today 👍 therefore I am currently working on issue 1056

server/services/invidious.ts
```diff
@@ -387,6 +440,99 @@
...
+		// 5b) Live-Probe gegen /api/v1/stats
+		let ok = false;
+		try {
+			const statsUrl = `https://${host}/api/v1/stats`;
+			const res = await this.api.get(statsUrl, {
+				headers: { Accept: "application/json" },
+				responseType: "json",
+				transformResponse: undefined,
+				validateStatus: s => s >= 200 && s < 500, // 2xx=ok; 4xx zählt als negativ, aber kein Throw
+			});
```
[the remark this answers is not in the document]
Sysadminfromhell: @dyc3 what does this mean? Whats the issue here?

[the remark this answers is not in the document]
Sysadminfromhell: on it.

Sysadminfromhell:
> Most of the logic you're implementing in `invidious.ts` should actually go in the info extractor.

Which parts actually?

dyc3:
After thinking about it for a while, I don't think this is the right direction to go in. It significantly increases the complexity of the info extractor, and we're just going to end up with SSRF and/or request amplification security issues.

Sorry about that. I think the better solution is to add some UI to let users tell the system that the link is an invidious/peertube link.

Sysadminfromhell:
> After thinking about it for a while, I don't think this is the right direction to go in. It significantly increases the complexity of the info extractor, and we're just going to end up with SSRF and/or request amplification security issues.
> 
> Sorry about that. I think the better solution is to add some UI to let users tell the system that the link is an invidious/peertube link.

I get that and I actually thought the same after this work. I feel like it would be great but I think to just give the admins/hosters the option to set allowed instances where they should know the api is available is better. I / we just need to documentate that correctly imo.

PR #1875, opened 2025-09-10
fix(server): Correctly handle a m3u8 manifest that contain segments directly

WuTofu:
This addresses a bug in the HLS service where the service would fail to parse m3u8 manifests that are media playlists (i.e., they contain segments directly) instead of master playlists (containing other playlists).

Test stream: https://test-streams.mux.dev/x36xhzz/url_6/193039199_mp4_h264_aac_hq_7.m3u8

### DIFF
--- a/server/services/hls.ts
+++ b/server/services/hls.ts
@@ -55,30 +55,38 @@ export default class HlsVideoAdapter extends ServiceAdapter {
 		const manifest = parser.manifest;
 		// log.silly(`Got m3u8 manifest with ${JSON.stringify(manifest)}`);
 
-		if (manifest.playlists.length === 0) {
-			throw new M3u8ParseError("No playlists found in manifest");
+		let duration = 0;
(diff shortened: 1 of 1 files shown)


Sysadminfromhell:
> This addresses a bug in the HLS service where the service would fail to parse m3u8 manifests that are media playlists (i.e., they contain segments directly) instead of master playlists (containing other playlists).
> 
> Test stream: https://test-streams.mux.dev/x36xhzz/url_6/193039199_mp4_h264_aac_hq_7.m3u8

Yes I can conform that behaviour. I had the same issue with the new odysee Adapter. If I resolve everything on the Server site and only give out the vx.m3u8 instead of the master.m3u8 it fails sometimes. I changed that behaviour on the ServerAdapter and called it "fixed" because, I thought, its an odysee thing.

dyc3 (approved):
Good work, thank you!

Issue #1872, opened 2025-09-09
Client: Chrome doesn't show subtitles in HLS Stream


Sysadminfromhell:
### Current Behavior

The CC/Subtitles Feature does not work on chrome as stated in #1866.

### Expected Behavior

Subtitles should be displayed as in Firefox

### Steps To Reproduce

Open Chrome
Visit offical OTT Site
Open a room
Pick a HLS Stream with CC/Subtitles copy the link and paste it into the add form and press play
Enable subtitles
See no captions

### Environment

- [x] This happens on the official site, opentogethertube.com
- [x] This happens using a self-hosted version.
- [x] I'm using the docker image.

PR #1878, opened 2025-09-11
fix(client): fix HLS captions menu and the captions not showing on chrome

WuTofu:
Fix issue where the HLS caption menu was not correctly displaying all available caption tracks listed in the manifest, and #1872

Test stream: https://storage.googleapis.com/shaka-demo-assets/angel-one-hls/hls.m3u8
(the first line of captions appears at approximately the 5 sec)

I also added some CSS to make the captions a bit responsive,, but I don't actually know anything about CSS. I'm open to any and all suggestions on how to improve it.

One thing I noticed was that without fix, captions were working on firefox, but when I entered PiP mode, captions were rendered inside the PiP windows, which is not what I expected and made me wonder if hls.js did something underlying to make the caption work on firefox.
With this fix, captions are now visible on both chrome and firefox, and the firefox's pip window no longer displays captions within it.

### DIFF
--- a/client/src/components/players/PlyrPlayer.vue
+++ b/client/src/components/players/PlyrPlayer.vue
@@ -118,9 +118,8 @@ export default defineComponent({
 					// (The SUBTITLE_TRACK_SWITCH event wasn't fired), it's unclear if this is a bug in Plyr or hls.js.
 					hls.subtitleTrack = trackIdx;
 
-					// Plyr Sometime switches back to -1 (disabled) after the track is set. We check again after a short delay.
-					// This appears to be a bug in plyr.js, as testing the same manifest on the hls.js
-					// demo page shows the captions track switching correctly.
+					// Plyr switches back to -1 (disabled) when the track's language code contains uppercase characters.
+					// So, we check again after a short delay. This appears to be a bug in plyr.js
 					setTimeout(() => {
 						if (player.value?.currentTrack === -1) {
 							console.error(


Sysadminfromhell:
@WuTofu would you mind fixing the lint errors?

Sysadminfromhell:
In my chrome there is still no Subtitles:

image

client/src/components/players/PlyrPlayer.vue
```diff
@@ -374,4 +375,9 @@ export default defineComponent({
 	object-fit: contain;
 	object-position: 50% 50%;
 }
+
+.plyr__captions {
+	font-size: clamp(1em, 3cqh, 3em);
+	bottom: 50px;
+}
```
dyc3: Let's avoid adding css to the plyr internals.

WuTofu:
@Sysadminfromhell I saw a cog icon in your screenshot. Could you please check if you are using my "hls-subtitles" branch?

Sysadminfromhell:
> @Sysadminfromhell I saw a cog icon in your screenshot. Could you please check if you are using my "hls-subtitles" branch?

Yea my mistake, switched branches now. Test and will come back to this pr.

client/src/components/players/PlyrPlayer.vue
Sysadminfromhell: @dyc3 I think imo to have the same visual settings set in the plyrplayer makes sense so the ux/ui for every browser looks the same. This change could benefit from "same experience everywhere". iirc every browser otherwise will interprete the styling differently when "left it unset" thats why I get crazy with fe work in the first place.

Sysadminfromhell:
@WuTofu, it would be more nice to know which setting represents which language it is. Use the attribute language instead of name it should be more generic. If youre fancy you could create a mapping for the languages to have a more / better ui experience. I also noticed, don't know if its just not available but in the xml, but stream 2 or 3 doesnt have any subtitles for me.

image


The Subtitles itself work on chrome but for my taste are a bit small:

image

client/src/components/players/PlyrPlayer.vue
WuTofu: Just took a quick look at plyr's README. What about something like this instead?
```css
.direct,
.plyr {
	--plyr-font-size-base: 3cqh;
	--plyr-font-size-small: 3cqh;
	--plyr-font-size-large: 3cqh;
	--plyr-font-size-xlarge: 3cqh;
}
```

Sysadminfromhell: I wouldnt go with 100% widht, we should stay in the recommended area between 75~95% somewhere there. When I remember my aborted PR for Dash CC/Subtitles support I set it to 92%, just gives us a bit of clearance left and right.

WuTofu:
@Sysadminfromhell Stream 2 and 3 were working perfectly for me earlier, but now I'm also experiencing this issue.
I can successfully download them using curl, so I guess we need to do some detective work to figure out what's wrong.

I'll focus on the caption switching functionality first. The font size also seems a bit too small to me, but I'll wait to see what dyc3 thinks about the method I proposed in my earlier reply to his comment.

client/src/components/players/PlyrPlayer.vue
WuTofu: @Sysadminfromhell What I proposed was "--plyr-font-size-*". To avoid confusion, I'll remove the existing code from my earlier reply.

WuTofu:
A quick update: there seems to be a bug in Plyr. In a very simple setup, switching to stream 3 immediately gets caption disabled. (https://codepen.io/wutofu/pen/MYadVza)

Currently, switching between streams 0-2 and "off" should be working fine.

Sysadminfromhell:
> A quick update: there seems to be a bug in Plyr. In a very simple setup, switching to stream 3 immediately gets caption disabled. (https://codepen.io/wutofu/pen/MYadVza)
> 
> Currently, switching between streams 0-2 and "off" should be working fine.

I might have an idea why, If we check the xml we see:

`#EXT-X-MEDIA:TYPE=SUBTITLES,URI="playlist_s-pt-BR.webvtt.m3u8",GROUP-ID="default-text-group",LANGUAGE="pt-BR",NAME="stream_3",AUTOSELECT=YES`

Even VLC Media Player has issue (doesn't show a language and needs long to load the subtitles).
I would assube it has somethin todo with the language code or the uri itself. Something is off here.

Sysadminfromhell:
New bug: if you set to stream_3 2x times (so once it get set to off and then once again) it will set to stream_0 but show selected stream_3:

Image

What I could observe is the following: Stream_3 sets the settrackindex to -1 which is not 0 (off) so it sets it to off BUT if you then move to stream_1 the index is 0. 
So something weird is happening. THis also happens with the latest hls 1.6.12. 
Therefore maybe we could test it further to see if it has something todo with the language "pt-BR" or is just a quirk of this specific file and call it a day.

<details>
  <summary>Console output</summary>

  ```
[log] > [subtitle-track-controller]: Switching to subtitle-track 0 "stream_0" lang:en group:default-text-group
Selected Track:  0
Selected Track:  -1
[log] > [subtitle-track-controller]: Switching to subtitle-track 0 "stream_0" lang:en group:default-text-group
  ```
</details>

WuTofu:
It looks like the language code was the problem. I downloaded the M3U8, replaced 'pt-BR' with 'pt,' and stream 3 worked perfectly.

I also took a look at the new bug you mentioned. It appears that stream 3 is somehow messing with Plyr's internal states, but since Plyr still fires the "languagechange" event, I've used that to update the caption's currentTrack.

Sysadminfromhell:
> It looks like the language code was the problem. I downloaded the M3U8, replaced 'pt-BR' with 'pt,' and stream 3 worked perfectly.
> 
> I also took a look at the new bug you mentioned. It appears that stream 3 is somehow messing with Plyr's internal states, but since Plyr still fires the "languagechange" event, I've used that to update the caption's currentTrack.

I have to differ, on the official demo site from hls I cannot reproduce this error with the file. There the subtitles work. It must be something "we" mess up. I assume its a plyrplayer issue then?

https://hlsjs-dev.video-dev.org/demo/

WuTofu:
> on the official demo site from hls I cannot reproduce this error with the file.

Yes, I noticed that as well. I mentioned this in my code's comment, but I forgot to mention it here

> It must be something "we" mess up.

I don't think that's the case (maybe?)
The issue also appears in [my simple test](https://codepen.io/wutofu/pen/MYadVza), which was built by following Plyr's hls.js example.

Sysadminfromhell:
> > It must be something "we" mess up.
> 
> I don't think that's the case (maybe?) The issue also appears in [my simple test](https://codepen.io/wutofu/pen/MYadVza), which was built by following Plyr's hls.js example.

I would assume that its the plyr's fault. I couldn't find any bug report of this issue. on HLS with a different player it works so I guess its something plyrplayer messes up at some point but I cant say what and why.

I guess I figured it out:

The player starts at -1 while the hls stream starts at 0 (always have one loaded), therefore the indexes are a mismatch. 
Enabling the CC w/o selecting one in the subsetting sets hls to 0 and player to -1, continue this we end at hls 2 and player 3 which does not work resulting in -1.

Here what I could see:

image

Setting it from 2 -> 3 temporary sets it to 3 just to rubberband to -1 again (off), idk why. but its a plyrplayer issue.

WuTofu:
> I would assume that its the plyr's fault. 

Just took some time looking into this, I believe [this line of code](https://github.com/sampotts/plyr/blob/0c9759455cbfcce888c66925c3b457ce06cee31e/src/js/captions.js#L122C19-L123C89) is the source of the issue. When plyr checks if a selected language exists, it's converted to lowercase and compared to stream 3's language code, which is "pt-**BR**".

I tried replacing "pt-BR" with "pt-br" in the m3u8 file, and then switching to and from stream 3 worked perfectly. I guess we can kick back a bit and call it a day :)

Sysadminfromhell:
> > I would assume that its the plyr's fault. 
> 
> 
> 
> Just took some time looking into this, I believe [this line of code](https://github.com/sampotts/plyr/blob/0c9759455cbfcce888c66925c3b457ce06cee31e/src/js/captions.js#L122C19-L123C89) is the source of the issue. When plyr checks if a selected language exists, it's converted to lowercase and compared to stream 3's language code, which is "pt-**BR**".
> 
> 
> 
> I tried replacing "pt-BR" with "pt-br" in the m3u8 file, and then switching to and from stream 3 worked perfectly. I guess we can kick back a bit and call it a day :)

Okay then we can "ignore" that for now, you could open a issue for plyrplayer because it's even in their documentation that they accept pt-BR. If not this might be an overseen issue on their site.

dyc3 (changes requested):
I tested it in chromium and could not get captions to show up using our test hls videos

WuTofu:
@dyc3, I just tested this in chromium 140.0.7339.127, and it worked fine for me. Could you please share your steps to reproduce this?

Also, in case you missed my conversation with Sysadminfromhell, we noticed that the language code for stream 3 contains uppercase characters. This could cause an issue with Plyr when switching to or from stream 3

dyc3:
Sorry for the long response time -- work has been ramping up recently and I can't dedicate as much time as I could previously.

I used chromium `140.0.7339.80`. I used OTT's test hls video "test hls 0" because that always has captions. I saw the captions menu, I clicked enable, and it showed a single caption for the current timestamp, and it didn't show anything after that.

WuTofu:
Thanks so much for the reply. You definitely don't need to apologize at all, I'm just happy to hear back from you when you have the time :)

I was actually able to reproduce the issue with the "test hls 0" . I'll dig into it later when I get a chance and see what's causing the problem.

WuTofu:
@dyc3, I didn't dig into it much, but this looks like another plyr issue. Since we don't use plyr's control interface anyway, what do you think about moving the HLS logic from PlyrPlayer.vue into a new component, HlsPlayer.vue, to handle HLS streams without plyr?

Sysadminfromhell:
> @dyc3, I didn't dig into it much, but this looks like another plyr issue. Since we don't use plyr's control interface anyway, what do you think about moving the HLS logic from PlyrPlayer.vue into a new component, HlsPlayer.vue, to handle HLS streams without plyr?

Does this fix the issue if we stick tho plyrplayer? We could change the backend instead to VideoJS. This should be available in the project already iirc.

dyc3:
The way it's set up is the way plyr recommends doing hls streaming. https://github.com/sampotts/plyr/issues/994 seems related, according to https://codepen.io/pen?template=oyLKQb

I suppose a new component could make sense, but it seems like a lot of work. I'd accept a new PR for it.

> We could change the backend instead to VideoJS.

I used video.js previously and it sucked. Integrating it into vue was hell, and it broke all the time.

WuTofu:
> We could change the backend instead to VideoJS. This should be available in the project already iirc.

@Sysadminfromhell, could you share your insight and elaborate more on this change? Since we're already using hls.js for hls streams in this project, and it's only plyr that's causing the caption issues, I would rather stick with hls.js and ditch plyr unless there's a specific reason not to.

> but it seems like a lot of work. I'd accept a new PR for it.

I'll see what I can do, but to be honest, typescript/vue isn't my strong suit. 😅

Sysadminfromhell:
> > We could change the backend instead to VideoJS. This should be available in the project already iirc.
> 
> 
> 
> @Sysadminfromhell, could you share your insight and elaborate more on this change? Since we're already using hls.js for hls streams in this project, and it's only plyr that's causing the caption issues, I would rather stick with hls.js and ditch plyr unless there's a specific reason not to.
> 
> 
> 
> > but it seems like a lot of work. I'd accept a new PR for it.
> 
> 
> 
> I'll see what I can do, but to be honest, typescript/vue isn't my strong suit. 😅

Hls.js handles the stream and PlyrPlayer displays it. We could change plyrplayer to VideoJS. But as @dyc3 explains: if it breaks too much I don't see it worth to replace. I mean we could try but as invidious also relay on this working flawless I don't see an option other as to open a big report to plyrplayer again and hope they can fix it. Or we find a way to patch that issues ourselves somehow.

WuTofu:
> Hls.js handles the stream and PlyrPlayer displays it

Assuming 'PlyrPlayer' refers to the plyr library, hls.js can actually not only handle the stream, but also display it. In my understanding, for HLS streams, plyr functions more as a control interface. That being said, hls.js also provides an API for controlling what the player displays, which is why I want to extract the HLS logic into a new component in the first place.

> invidious also relay on this working flawless

I just took a quick look at #1862. It looks like it was just a server-side change, so it should be fine, but we'll see! Hopefully this doesn't create a lot of extra work on you.

WuTofu:
Closing this PR, as the work is being continued in #1880

PR #1885, opened 2025-09-19
client: enhance subtitle positioning based on player controls visibility

WuTofu:
This PR improves subtitle display in the video player by dynamically adjusting their vertical position depending on whether the player controls are visible. When controls are shown, subtitles are nudged higher to avoid overlap; when controls are hidden, subtitles move lower for better readability.

Note: This adjustment uses css targeting webkit-based browsers (chrome, edge, safari) and will not work on Firefox due to lack of support for ::-webkit-media-text-track-display.

### DIFF
--- a/client/src/views/Room.vue
+++ b/client/src/views/Room.vue
@@ -27,7 +27,13 @@
 			</div>
 			<div class="video-container">
 				<div class="video-subcontainer">
-					<div class="player-container" :class="{ 'controls-visible': controlsVisible }">
+					<div
+						class="player-container"
+						:class="{
+							'controls-visible': controlsVisible,
+							'controls-just-hidden': controlsJustHidden,
+						}"
+					>
 						<OmniPlayer
 							:source="store.state.room.currentSource"
 							@apiready="onPlayerApiReady"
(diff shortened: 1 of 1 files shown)


dyc3:
I was testing this using "test hls 0", and I was seeing about half of the captions fly in from the top, and some that were completely unaffected.

Some searching led me to find the `::cue` pseudo element: 
- https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Audio_and_video_delivery/Adding_captions_and_subtitles_to_HTML5_video#styling_the_displayed_subtitles
- https://developer.mozilla.org/en-US/docs/Web/CSS/::cue

WuTofu:
> I was testing this using "test hls 0", and I was seeing about half of the captions fly in from the top, and some that were completely unaffected.

Thanks for testing that out! It looks like that "transition" I added is contributing to the captions flying in from the top. I'll try to get it working as expected.

> Some searching led me to find the ::cue pseudo element

Yeah, I also found that pseudo element, but when I opened the first link in firefox, I saw the css style wasn't working for the demo. Also, it looks like ::cue supports only a few CSS properties, and those don't include "bottom" or "transform".

WuTofu:
Well, I have a fix for the captions flying in from the top, but then I noticed some captions were shifted outside the player's visible area.

Initially, I thought they were captions with "line: 0%" as one of their cue settings, but I was wrong, and I wondered how they got themselves on top of the player.
For example, a caption in "English" track (not the forced one) at 1:47 shows: "NOW MY NAME APPEARS AT THE BOTTOM OF THE SCREEN".
But when I checked the vtt file, I couldn't find a line like that in there.

@dyc3, do you have any idea? :(

dyc3:
Nope, sorry, this is new territory for me too.

WuTofu:
Okay, I figured it out!

So, I was wondering if the captions weren't in the vtt file, could they be embedded in the video stream instead? That's when I remembered that hls.js supports CEA-608/708 captions. Sure enough, setting 'enableCEA708Captions' to false in a new hls instance made the line of caption I mentioned earlier disappear.

I guess there is a way if we really want to manipulate the position of all kinds of subtitles, but I don't want to overcomplicate how the player displays them.

Closing this PR until I have a better idea.

PR #1886, opened 2025-09-20
chore: improve HlsPlayer component structure

WuTofu:
No new fixes or features here. I just focused on improving the code structure.

My apologies for the previous messy version. 🤦‍♂️

### DIFF
--- a/client/src/components/players/HlsPlayer.vue
+++ b/client/src/components/players/HlsPlayer.vue
@@ -5,15 +5,15 @@
 			preload="auto"
 			crossorigin="anonymous"
 			:poster="thumbnail || ''"
-			@canplay="onReady"
-			@ready="onReady"
-			@playing="onPlaying"
-			@pause="onPaused"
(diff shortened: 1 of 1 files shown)


dyc3:
I actually prefer the way it is now, I don't think this change is necessary :)

WuTofu:
Sure, I'll leave it as is if you prefer.

Off topic for a sec, from my understanding, reddit also uses hls, right?
https://github.com/dyc3/opentogethertube/blob/c5dc255ecdfe8f9c76a881bca2d470c32b20c85f/server/services/reddit.ts#L133-L142

dyc3:
Yes, but reddit has been disabled for quite some time now. Ever since they took away free access to their API.

WuTofu:
Ah, I see. That explains why the Reddit videos are working on my instance, but not on the official site.

I'll open a PR to use the new HlsPlayer.vue to play them.

PR #1887, opened 2025-09-20
Server: Add Field filtering for YouTube API Calls

### LINKED ISSUE #211: Use fields parameter to reduce size of youtube api responses
The youtube API has a `fields` parameter that allows you to filter only for the stuff you need. We don't use a significant portion of the data we get from the API, so this seems like a reasonable optimization.

https://developers.google.com/youtube/v3/getting-started#fields

Sysadminfromhell:
This PR should close #211.

Im my test the api call was 0.02 seconds faster.

### DIFF
--- a/server/services/youtube.ts
+++ b/server/services/youtube.ts
@@ -296,6 +296,7 @@ export default class YouTubeAdapter extends ServiceAdapter {
 					key: this.apiKey,
 					part: "contentDetails",
 					[channelIdProp]: channelIdValue,
+					fields: "items(id,contentDetails(relatedPlaylists(uploads)))",
 				},
 			});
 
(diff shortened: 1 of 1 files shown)


server/services/youtube.ts
```diff
@@ -672,7 +698,12 @@ export default class YouTubeAdapter extends ServiceAdapter {
 		};
 
 		try {
-			const res = await this.api.get("/search", { params });
+			const res = await this.api.get("/search", {
+				params: {
+					...params,
+					fields: "items(id(videoId))",
+				},
```
dyc3: Why not just add it to `params`?

server/services/youtube.ts
```diff
@@ -533,6 +540,25 @@ export default class YouTubeAdapter extends ServiceAdapter {
 		}
 	}
 
+	private getVideosFieldsParam(parts: YoutubeApiPart[]): string {
```
dyc3: nit: fits existing semantics better

```suggestion
	private getNeededFields(parts: YoutubeApiPart[]): string {
```

server/services/youtube.ts
Sysadminfromhell: I forgot to move it to params. I put it there for testing.

PR #1890, opened 2025-09-20
An attempt to fix the e2e test

WuTofu:
I didn't actually know much about cypress e2e tests, but it kind of annoyed me to see that test fail, so I took an educated guess and jumped the gun. I hope you don't mind.


WuTofu:
Now we are having this:

> We detected that the Electron Renderer process just crashed.
> 
> We have failed the current spec but will continue running the next spec.
> 
> This can happen for a number of different reasons.
> 
> If you're running lots of tests on a memory intense application.
>   - Try increasing the CPU/memory on the machine you're running on.
>   - Try enabling experimentalMemoryManagement in your config file.
>   - Try lowering numTestsKeptInMemory in your config file during 'cypress open'.

Will check if there is anything I can do when I have a moment

dyc3:
> We detected that the Electron Renderer process just crashed.

This has been happening forever and it is unrelated. I'm pretty sure its Cypress's fault because it says:
```
Terminating renderer for bad IPC message, reason 114
```

The playback e2e tests are super flaky, and it would be a HUGE improvement to have that fixed.

WuTofu:
@dyc3 After some searching, it did feel like it took forever for cypress to fix this.

However, it also led me to this thread [1], and I decided to give cypress 15 a shot.
I checked the migration guide for cypress 14 and 15. It looked like this should be a pain-free upgrade for us.
The only issue is that cypress 15 no longer supports node 18, but luckily, we just upgraded the node version for this project to 20 recently. :)

I ran the e2e tests [2] 10 times in a row, they all passed. The 9th and 10th runs included the change with the null check in chat.vue to fix the flaky chat test.

[1] https://github.com/cypress-io/cypress/issues/27415#issuecomment-3266391302
[2] https://github.com/WuTofu/opentogethertube/actions/workflows/e2e.yml

PR #1893, opened 2025-09-24
fix(server): Sequelize-cli now respects env files

WuTofu:
This pr updates server/config/config.mjs for

1. TOML parsing
2. Fix typo? (Replaced the use of conf.load with conf.loadFile)

### DIFF
--- a/server/config/config.mjs
+++ b/server/config/config.mjs
@@ -78,13 +78,13 @@ const conf = convict({
   },
 });
 try {
-  conf.load("../env/base.toml");
+  conf.loadFile("../env/base.toml");
 } catch {
   console.log("No base config found");
 }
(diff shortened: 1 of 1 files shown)


dyc3:
I need a bit more context, what is this fixing?

WuTofu:
Well, I encountered an issue related to using sqlite in our docker (prod) environment

Our docker image always executes `wait_for_db.sh` with `NODE_ENV` set to `production` to run the database migrations. As a result, without this fix, even though I had already set db.mode in production.env to "sqlite", it would fail to connect to the postgres database.
https://github.com/dyc3/opentogethertube/blob/60697ec809a8b2fe3b953e1ac50af08ba14aebc2/deploy/monolith.Dockerfile#L51

PR #1894, opened 2025-09-27
server: Replace "secure-password" with "argon2"

### LINKED ISSUE #1870: Replace `secure-password` dependency with something else.
`secure-password` is unmaintained., and recently caused problems for us when upgrading to node 20. It will likely continue to be a problem for future version upgrades.

So far `node-argon2` looks like it will make the most sense.

WuTofu:
Close #1870

I checked the earliest version of `secure-password` that was used in this project, which is v3.1.0.
https://github.com/dyc3/opentogethertube/blob/1b9b03041dd6c001cd218011bec59341c27f1c5c/server/package.json#L39

This is great news! That version of secure-password checked [1] whether the input had `$argon2i$` or `$argon2id$`, meaning it was already using argon2. So, we should be able to just replace the library and slightly adapt the password checking logic to `node-argon2`'s API, and call it a day :D

In my dev environment, I created an account on the server using `secure-password`, and then restarted the server to use `node-argon2` and then I could successfully login to my test account.
I also created a second test account on the server using `node-argon2`, then logged off, and could successfully log back in.

I didn't test the account recovery flow, it seems that it is part of the e2e integration test, right?

[1] https://github.com/emilbayes/secure-password/blob/275e4634e2ac119ad885c3625a77ae7217c02435/index.js#L128


dyc3:
Did you test this with both sqlite and postgres?

WuTofu:
Yes, I did.

I tested this with sqlite in my dev environment, and with postgres in my prod (docker) environment

dyc3 (approved):
Since the e2e tests pass, I'm pretty confident this works as intended. However, I would appreciate at least a unit test that makes sure that a hash (hard coded in the test) generated before this change works after this change.

WuTofu:
> a unit test that makes sure that a hash (hard coded in the test) generated before this change works after this change

I've added the requested unit test. Could you please review and see if this is what you want?
Please let me know if anything looks off or can be improved :)

dyc3 (approved):
Yup, looks good! Thank you!

PR #1903, opened 2025-10-03
fix(client): Use useMouseInElement instead to trigger video controls during playback

WuTofu:
Only mouse movement within the `player-container` will trigger the video controls, improving user experience during playback.

### DIFF
--- a/server/tests/unit/services/peertube.spec.ts
+++ b/server/tests/unit/services/peertube.spec.ts
@@ -16,10 +16,13 @@ import { AxiosRequestHeaders, AxiosResponse } from "axios";
 import { conf, loadConfigFile } from "../../../ott-config.js";
 
 const validVideoLinks = [
-	["the.jokertv.eu:7C5YZTLVudL4FLN4JmVvnA", "https://the.jokertv.eu/w/7C5YZTLVudL4FLN4JmVvnA"],
+	[
+		"tube.shanti.cafe:96kFpg7fs4LwYkFCMFv51E",
+		"https://tube.shanti.cafe/w/96kFpg7fs4LwYkFCMFv51E",
+	],
 ];
 
-const invalidLinks = ["https://the.jokertv.eu/", "https://the.jokertv.eu/w/"];
+const invalidLinks = ["https://tube.shanti.cafe/", "https://tube.shanti.cafe/w/"];
 
 const FIXTURE_DIRECTORY = "./tests/unit/fixtures/services/peertube";
 


dyc3:
The reason I hadn't implemented this before was because the logic for detecting mouse movements over iframes was very flaky. The current behavior (mouse movements anywhere activates the controls) provides a reasonable fallback that is completely seamless to most users.

Does this still work when the mouse is over an iframe? And when the iframe has focus?

WuTofu:
Ah, to be honest, I wasn't aware that iframes could be an issue.

Currently, it seems only YouTube and PeerTube might use iframes, but YouTube's video control is always displayed.
I tried several peertube links, but they all use hls. ~Do you have a peertube link that still uses iframe that I could test with?~

Edit: Nevermind. Just realized that I should config the info_extractor to use peertube iframe.

WuTofu:
Yeah, you are right. When the iframe has focus, this doesn't work. I will update the code

client/src/views/Room.vue
```diff
@@ -292,11 +294,22 @@ export default defineComponent({
 			}
 		}
 
+		const isIframeBasedPlayer = computed(() => {
+			const service = store.state.room.currentSource?.service;
+			return ["youtube", "peertube"].includes(service || "");
+		});
```
dyc3: This feels rather brittle. It's assuming the implementation details of these players. Could be a footgun later.

WuTofu:
> I actually don't think the iframe player check is technically necessary.

I added that check because I wasn't able to activate the controls for the iframe player after you asked me if this still worked when the mouse was over an iframe that had focus.

And I just tried again without the iframe player check, surprisingly, the controls could be activated, is this the flakiness you were talking about? :(

dyc3:
> is this the flakiness you were talking about? :(

Hmm, yeah it is. Could be a browser thing, I was using firefox

WuTofu:
Well, after  a couple of tests, I once again encountered the problem where the controls could not be activated when the iframe player like peertude had a long loading time while it had focus. However, it is not always reproducible, so I will keep the iframe check to maintain the current behavior for iframe players

client/src/components/AddPreview.vue
```diff
@@ -156,7 +156,7 @@ const testVideos = import.meta.env.DEV
 				"test dash 3 w/ multiple captions",
 				"https://livesim2.dashif.org/vod/testpic_2s/multi_subs.mpd",
 			],
-			["test peertube 0", "https://the.jokertv.eu/w/7C5YZTLVudL4FLN4JmVvnA"],
+			["test peertube 0", "https://tube.shanti.cafe/w/96kFpg7fs4LwYkFCMFv51E"],
```
dyc3, line 163: I would also accept a change to replace `the.jokertv.eu` in the default config with a different instance :)

dyc3:
I'll give this another thorough review sometime soon.

client/src/components/AddPreview.vue
WuTofu: I've updated the default instances list with a selection from https://instances.joinpeertube.org/ :)

dyc3:
There's a unit test that also needs to be updated, see CI

WuTofu:
@dyc3, apologies, I got sidetracked right after pushing that change, and forgot to get back and check the CI 😅

PR #1908, opened 2025-10-07
refactor(client): simplify closed captions switcher and introduce video settings menu

### LINKED ISSUE #1001: make subtitles button be a toggle

When you click on the subtitles button, it asks you "on" or "off"

it should be a toggle, like it is on youtube

### Additional context

_No response_

WuTofu:
This pr refactors the video player controls by consolidating the closed captions and video quality switchers into a single, unified settings menu.

1. The new `VideoSettings` component allows users to manage subtitles and video quality from one place.
2. Removed the old `VideoQualitySwitcher` 
3. Updated the localization strings.
4. Close #1001

### DIFF
--- a/tests/e2e/integration/playback.spec.ts
+++ b/tests/e2e/integration/playback.spec.ts
@@ -108,6 +108,8 @@ describe("Video playback", () => {
 		// should have both captions and playback rate controls enabled
 		cy.get('[aria-label="Playback Speed"]').should("exist").should("be.enabled");
 		cy.get('[aria-label="Closed Captions"]').should("exist").should("be.enabled");
+		// the settings button should always be enabled, even if there's nothing to set
+		cy.get('[aria-label="Player settings"]').should("exist").should("be.enabled");
 		cy.get("video").should("exist").scrollIntoView();
 
 		// change the playback rate to 1.5x
@@ -116,7 +118,7 @@ describe("Video playback", () => {
 		cy.get("video").should(element => {
 			expect(element[0].playbackRate).to.be.equal(1.5);
 		})
-		
+
 		// // FIXME: change the volume to 10%
 		// cy.get('[data-cy="volume-slider"]').ottSliderMove(10);
 		// cy.get("video").should(element => {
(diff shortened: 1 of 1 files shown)


WuTofu:
Ah, forgot to update the e2e test to reflect this change.

@dyc3, could you please take a look and let me know your thoughts on this pr before I proceed?

client/src/components/controls/VideoSettings.vue
```diff
@@ -0,0 +1,316 @@
...
+				<transition name="menu-resize" mode="out-in" slim>
+					<!-- Main menu -->
+					<v-list
+						v-if="currentMenu === MenuType.MAIN"
+						key="main"
+						class="menu-content"
+						min-width="300px"
+					>
+						<v-list-item
+							v-if="isCaptionsSupported"
```
dyc3: We should show it disabled when it's not supported.

client/src/components/controls/VideoSettings.vue
```diff
@@ -0,0 +1,316 @@
...
+							<div class="menu-item-content">
+								<span>{{ $t("room.subtitles") }}</span>
+								<span v-if="currentSubtitleDisplay" class="menu-item-value">
+									{{ currentSubtitleDisplay }}
+								</span>
+							</div>
+						</v-list-item>
+
+						<v-list-item
+							v-if="isQualitySupported"
```
dyc3: Same here

client/src/components/controls/VideoSettings.vue
```diff
@@ -0,0 +1,316 @@
...
+.video-settings-wrapper {
+	position: relative;
+}
+
+.settings-menu-container {
+	position: absolute;
+	bottom: media-controls.$video-controls-height;
+	right: -90px;
+	z-index: 9999;
+	background: rgba(0, 0, 0, 0.7);
```
dyc3: The background shouldn't be transparent. It also doesn't respect theme colors.

This is with the strawberry theme:

image

image

client/src/components/controls/VideoSettings.vue
```diff
@@ -0,0 +1,316 @@
...
+	border-radius: 10px;
+	padding: 0;
+	width: auto;
+	box-shadow: 0 4px 20px rgba(0, 0, 0, 0.3);
```
dyc3: box shadow should also respect theme colors

client/src/components/controls/VideoSettings.vue
WuTofu: @dyc3, Don't you think showing the caption option as disabled when it’s not supported might be misleading? It could imply that users can enable it.

Technically, I could let it show "disabled" and use `!isCaptionsSupported` in `:disable` to make it unclickable. But if we do want to show the caption row regardless of support, I’m wondering if there’s a clearer way to word it.

client/src/components/controls/VideoSettings.vue
WuTofu: I will see what I can do with the theme colors.

@dyc3, since this menu appears at the top of the video, don't you think a transparent background might give it a more modern and visually appealing look? :(

client/src/components/controls/VideoSettings.vue
dyc3: It communicates to the user that "the feature is here, but you can't use this right now", without the user thinking "where's the captions? was it here? is it in preferences?" etc.

client/src/components/controls/VideoSettings.vue
dyc3: Legibility is my base concern. I'm not entirely opposed to having it be a little transparent, but it probably should be 90% opacity instead of 70%.

WuTofu:
@dyc3, I've implemented your feedback and applied the transparent background and rounded menu style to the playback rate switch as well

PR #1915, opened 2025-10-11
feat(client): improves video quality labeling

WuTofu:
This PR improves video quality labeling in the quality menu by switching from using only video height to using both width and height for video track identification and display. 

This introduces a new function to generate user-friendly resolution labels based on aspect ratio and resolution, and adds unit tests to ensure correct behavior across a wide range of video formats and edge cases.

### DIFF
--- a/client/src/util/misc.ts
+++ b/client/src/util/misc.ts
@@ -9,7 +9,12 @@ export function isOfficialSite(): boolean {
 }
 
 function gcd(a: number, b: number): number {
-	return b === 0 ? a : gcd(b, a % b);
+	while (b !== 0) {
+		const temp = b;
+		b = a % b;
+		a = temp;
+	}
+	return a;
 }
 
 /**


client/src/util/misc.ts
```diff
@@ -5,3 +5,154 @@ export function enumKeys<O extends object, K extends keyof O = keyof O>(obj: O):
 export function isOfficialSite(): boolean {
 	return window.location.hostname.endsWith("opentogethertube.com");
 }
+
+function gcd(a: number, b: number): number {
+	return b === 0 ? a : gcd(b, a % b);
+}
```
dyc3, line 18: Would prefer this to not be recursive

client/src/components/composables/media-player.ts
```diff
@@ -52,7 +52,7 @@ export interface MediaPlayerWithPlaybackRate extends MediaPlayer {
 }
 
 export interface MediaPlayerWithQuality extends MediaPlayer {
-	getVideoTracks(): number[];
+	getVideoTracks(): { width: number; height: number }[];
```
dyc3: We should go ahead and make `{ width: number; height: number }` it's own named type.

PR #1916, opened 2025-10-12
feat(client): Add visual indication for caption tracks

WuTofu:
This PR includes the following changes:

1. Changed the argument type for setCaptionsTrack from string to number, allowing us to switch between text tracks using an index
2. Introduced the `CaptionTrack` interface in `media-tracks.ts` to represent caption/subtitle tracks.
3. Added an icon to the text track selection menu to indicate whether a track is a caption track

image

### DIFF
--- a/client/src/components/controls/VideoSettings.vue
+++ b/client/src/components/controls/VideoSettings.vue
@@ -168,7 +168,13 @@ const currentSubtitleDisplay = computed(() => {
 });
 
 function formatCaption(track: CaptionTrack): string {
-	return track.label || track.srclang || "unknown";
+	const localiedLabel =
+		track.srclang &&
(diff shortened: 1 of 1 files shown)


WuTofu:
@Sysadminfromhell once mentioned this in https://github.com/dyc3/opentogethertube/pull/1878#issuecomment-3281077973
> If youre fancy you could create a mapping for the languages to have a more / better ui experience.

A quick search led me to `Intl.DisplayNames()`[1] which seems to be an easy way to implement this feature without introducing a new dependency or requiring lots of code changes.

Now the menu looks like this:
image

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames/DisplayNames

PR #1917, opened 2025-10-13
chore(client): Enhance test video display with grouped categories and improved styling

WuTofu:
While it's probably not necessary to improve this test video selection menu since it's only for devs, sometimes we just want to have something nice for ourselves. 😉

image

### DIFF
--- a/client/src/components/AddPreview.vue
+++ b/client/src/components/AddPreview.vue
@@ -25,7 +25,7 @@
 			<v-container v-if="!production" class="test-videos-container">
 				<v-row>
 					<v-col
-						v-for="(group, groupName) in groupedTestVideos"
+						v-for="(group, groupName) in testVideos"
 						:key="groupName"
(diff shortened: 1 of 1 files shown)


client/src/components/AddPreview.vue
```diff
@@ -196,6 +213,45 @@ const production = computed(() => {
...
+		} else if (label.includes("dash")) {
+			groups.DASH.push(video);
+		} else if (label.includes("peertube")) {
+			groups.PeerTube.push(video);
+		} else if (label.includes("odysee")) {
+			groups.Odysee.push(video);
+		} else {
+			groups.MISC.push(video);
+		}
+	});
```
dyc3: I would rather just restructure the test videos to include the necessary metadata. Ideally this function isn't necessary at all.

WuTofu: Done 👍

PR #1920, opened 2025-10-14
refactor(server/client): simplify player selection logic

WuTofu:
This pull request refactors how different video sources are handled throughout both the client and server codebases. 

1. Changed the service ID returned by several adapters (`GoogleDriveAdapter`, `OdyseeAdapter`, `PlutoAdapter`, `RedditAdapter`, `TubiAdapter`) to use `direct` or `hls` instead of their previous custom values. (Kind of like peertube, we only use the custom value when we use its iframe player or need some specifial handling) This simplifies frontend logic for selecting the appropriate player component.
2. Removed the `GoogleDrivePlayer.vue` component and its usage from `OmniPlayer.vue`, consolidating google drive playback under the generic direct-play player.
3. Removed the `video.js` from the client dependencies since `GoogleDrivePlayer.vue` was removed.


WuTofu:
I'll work on the unit test if you think this PR can move forward.

dyc3 (changes requested):
While this hasn't really been a strict rule, the original intent of the `source` field in the `Video` type was to describe where a video came from, not necessarily the player it will use. But in practice, `source` became "what player will be used." 

I do like getting rid of videojs entirely. Forgot we still had it.

I think the better solution is to decouple video source and video player, but that's a rather large task. While this does technically simplify the logic, it muddies the water for what `source` means.

WuTofu:
> I think the better solution is to decouple the video source and the video player

Hmm, I'm not 100% sure I understand what you mean, like how do you want the video source and video player decoupled?

I initially thought we could use only one of the `service` or `mime` field to determine the player on the frontend, but since the server already checks `canHandleURL`, that might be redundant. Would introducing a new field like `player` in the `Video` type align with what you want?

dyc3:
> Would introducing a new field like player in the Video type align with what you want?

I would be open to separating "video metadata" (aka info about the video) and "how to play the video". If what you suggest makes the most sense, then sure. However, a change like this is rather fundamental. It requires a bit more diligence to pull off without incident.

WuTofu:
> If what you suggest makes the most sense, then sure.

I haven’t had a chance to study how the server works thoroughly yet, so I probably need to dig into it before planning my next steps.

> However, a change like this is rather fundamental. It requires a bit more diligence to pull off without incident.

Yeah, I agree. To keep things clean, I will move the changes to `GoogleDrivePlayer.vue` and `video.js` into a separate PR

Sysadminfromhell:
I would also like to add: If we change the service to for example hls the video gets processed 2x times. Once my Info extractor trying to get the video info (for exmaple odysee because its an odysee link) when this returns hls it will try to get the video when played via hls. Which can, in case of odysee, sometimes wont work how intended. We need further testing in any effected service then. I have the same issue with Invidious, it uses first Invidious.ts then switches, depending on the return value of the service, a second time through the service-meant.ts.
This is not needed due to both APIs would not need to be examine by 2 different Service Adapter files because the length in both cases can be given over cleanly. I worked around this issue in invidious but got rid off it entirely in odysee. Set the log level to debug and see what I mean @WuTofu.

WuTofu:
> If we change the service to for example hls the video gets processed 2x times

> when this returns hls it will try to get the video when played via hls. Which can, in case of odysee, sometimes wont work how intended.

>  I have the same issue with Invidious, it uses first Invidious.ts then switches, depending on the return value of the service, a second time through the service-meant.ts.

> Set the log level to debug and see what I mean

After setting it to debug, what else should I do to see this "processing 2x" issue?

Sysadminfromhell:
> > If we change the service to for example hls the video gets processed 2x times
> 
> > when this returns hls it will try to get the video when played via hls. Which can, in case of odysee, sometimes wont work how intended.
> 
> > I have the same issue with Invidious, it uses first Invidious.ts then switches, depending on the return value of the service, a second time through the service-meant.ts.
> 
> > Set the log level to debug and see what I mean
> 
> After setting it to debug, what else should I do to see this "processing 2x" issue?

Sorry I didnt see your comment. 

You should see that the video is processed by the infoextractor 2x times instead of one. This can be fixed with a custom type for invidious. Invidious now sets the type accodingly to the video format it gets from youtube. the first infoextractor "round" then determinds its invidious and sends it to this pipe, out comes a dash video whch then gets processed by dash. I "used" this method to not modifiy too much of the core content but with the experience I have now I would change invidious to have its own working pipe and infoextractor does not need to process this video twiece if invidious already gave all the infos.

PR #1921, opened 2025-10-15
feat(client): display an error message when video is failed to load

### LINKED ISSUE #1919: HLS/DASH players don't display any errors to the user when network errors occur
### Current Behavior

_No response_

### Expected Behavior

If the request for the video fails, the player show an error to the user.

### Steps To Reproduce

1. use ffmpeg to create an HLS stream from a video
2. serve the video using an http server, like `miniserve`
3. Add an hls video to the queue
4. Make sure the server never includes any CORS headers
5. See that video is not actually loading, and the js console indicates network errors

### Environment

- [x] This happens on the official site, opentogethertube.com
- [x] This happens using a self-hosted version.

### Room name or URL

_No response_

### Video URL

_No response_

### Anything else?

_No response_

WuTofu:
Close #1919.

@dyc3, could you take a look and let me know if this looks good or anything you think could be improved when you have a moment?

image

### DIFF
--- a/client/src/components/players/DashPlayer.vue
+++ b/client/src/components/players/DashPlayer.vue
@@ -12,7 +12,6 @@
 			@stalled="onBuffering"
 			@loadstart="onBuffering"
 			@progress="onProgress"
-			@error="onError"
 		></video>
 	</div>
 	<div id="dashplayer-ttml-rendering"></div>
@@ -20,8 +19,9 @@
 
 <script lang="ts" setup>
 import { onMounted, ref, watch, onBeforeUnmount, toRefs } from "vue";
-import { MediaPlayer, type MediaPlayerClass } from "dashjs";
+import { MediaPlayer, type MediaPlayerClass, type ErrorEvent } from "dashjs";
 import type {
+	MediaPlayerError,
 	MediaPlayerWithCaptions,
 	MediaPlayerWithPlaybackRate,
 	MediaPlayerWithQuality,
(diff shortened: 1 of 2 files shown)


client/src/components/composables/media-player.ts
```diff
@@ -176,3 +176,8 @@ export function usePlaybackRate() {
 		availablePlaybackRates,
 	};
 }
+
+export enum MediaPlayerError {
+	none = 0,
+	network = 1,
+}
```
dyc3: Would prefer this to not be a ts enum. Should be a union instead.

I know this isn't consistent with the rest of the codebase, but I've changed my opinion on ts enums since then.

client/src/components/composables/media-player.ts
```diff
@@ -176,3 +176,8 @@ export function usePlaybackRate() {
 		availablePlaybackRates,
 	};
 }
+
+export enum MediaPlayerError {
+	none = 0,
+	network = 1,
```
dyc3: If there is no media player error, then it should that should be represented with `null` or `undefined`, not as a part of the error.

In other words, if its an error, its some value. If there's no error, its null/undef.

client/src/components/players/OmniPlayer.vue
```diff
@@ -409,4 +437,17 @@ const renderedSpans = computed(() => {
...
+	width: 100%;
+	height: 100%;
+	align-content: center;
+	background-color: rgba(var(--v-theme-surface), 1);
+	z-index: 1;
+
+	.playback-error-text {
+		margin-left: 20%;
+		max-width: 50%;
+	}
```
dyc3: Would rather this be centered, prefer flex layout.

client/src/components/players/OmniPlayer.vue
```diff
@@ -409,4 +437,17 @@ const renderedSpans = computed(() => {
 	font-size: 12px;
 	z-index: 500;
 }
+.playback-error {
+	position: absolute;
+	width: 100%;
+	height: 100%;
+	align-content: center;
+	background-color: rgba(var(--v-theme-surface), 1);
```
dyc3: I think its more consistent if we use the background color, not the surface color

client/src/locales/en.ts
```diff
@@ -406,5 +407,15 @@ export default {
...
 		},
+		"playback-error-title": {
+			[MediaPlayerError.none]: "No Error",
+			[MediaPlayerError.network]: "Playback Error",
+		},
+		"playback-error-message": {
+			[MediaPlayerError.none]:
+				"No error has occurred. If you're seeing this, please report it as a bug.",
+			[MediaPlayerError.network]:
+				"Failed to load video. This is most likely due to the video being blocked by CORS policy, or the network issue. If you believe this is an error, please report it as a bug.",
```
dyc3: We should have info available to provide a more fine grained error message. We shouldn't suggest it's a CORS issue unless we know. It could be that hls.js failed to parse the incoming segment.

I would be fine with just displaying the raw error message for now. It matters more that we display something actionable (even if its a little technical), rather than lead users in the wrong direction.

PR #1924, opened 2025-10-17
client: only show the PiP button when browser supports the PiP API

WuTofu:
Just noticed that firefox doesn't even support the picture-in-picture API and since it provides a native PiP button on videos, our custom button is redundant. So, hide this useless button from firefox users


WuTofu:
@dyc3, I have updated the code. Not sure why the re-request review button on  github is gone for me right now :(

PR #1925, opened 2025-10-17
client: add a volume icon and improve volume slider's tooltip style

WuTofu:
While it's not necessary to have a volume icon, it does feel like something is missing when I look at the volume slider.
I also added `isMuted` and `prevVolumne` to track mute status and previous volume in `media-player.ts`, and implemented logic to synchronize mute and volume changes, including restoring previous volume when unmuting

I also added some volume slider UI enhancements:
1. Positioned the tooltip below the slider, which matches how we position other tooltips.
2. Updated the tooltip style as the theme changes
3. Made the tooltip go away when the mouse moves away. Instead of requiring the user to click anywhere on the page to make it disappear.

### DIFF
--- a/client/src/components/players/OmniPlayer.vue
+++ b/client/src/components/players/OmniPlayer.vue
@@ -268,7 +268,7 @@ async function onApiReady() {
 	captions.isCaptionsSupported.value = isCaptionsSupported();
 	qualities.isQualitySupported.value = isQualitySupported();
 	if (player.value) {
-		player.value.setVolume(volume.value);
+		player.value.setVolume(volume.volume.value);
 	}
 	if (implementsCaptions(player.value)) {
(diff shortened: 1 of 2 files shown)


WuTofu:
@dyc3, just curious, any reason that you didn't use the `v-slider` when you implemented this? Or was it just not available at that time?

[1] https://vuetifyjs.com/en/components/sliders/

dyc3:
Was easier to remain visually consistent at the time

WuTofu:
> Was easier to remain visually consistent at the time

I see. Would you mind if I try to replace `vue-slider` with `v-slider`?

dyc3 (approved):

dyc3:
Looks like the e2e tests need to be updated.

> Would you mind if I try to replace vue-slider with v-slider?

I would rather keep it as is for now.

PR #1950, opened 2026-02-15
fix: reject newlines in room title validation

### LINKED ISSUE #555: Server should not accept room title or name with newlines
### Current Behavior

It currently possible for a room title to contain a newline, even though the UI does not allow it.

### Expected Behavior

`PATCH /room/:roomName` should not allow 

### Steps To Reproduce

Send `PATCH /room/:roomName` with a title that contains a newline

### Environment

### Anything else?

_No response_

ugurcl:
## Summary
- Add newline and carriage return validation to room title field in Zod schemas
- Applies to both room creation and room settings update
- The name field is already protected by ROOM_NAME_REGEX which only allows [a-z0-9_-]

## Changes
- common/models/zod-schemas.ts - Added .refine() to reject newlines in title for both create and patch schemas
- server/tests/unit/api/room.spec.ts - Added test cases for \n and \r in title for both POST and PATCH endpoints

Closes #555

### DIFF
--- a/common/models/zod-schemas.ts
+++ b/common/models/zod-schemas.ts
@@ -30,7 +30,11 @@ export const OttApiRequestRoomCreateSchema = z
 			.refine(name => !RESERVED_ROOM_NAMES.includes(name), {
 				message: "not allowed (reserved)",
 			}),
-		title: z.string().max(255, "too long, must be at most 255 characters").optional(),
+		title: z
+			.string()
+			.max(255, "too long, must be at most 255 characters")
+			.refine(s => !/[\n\r]/.test(s), { message: "must not contain newlines" })
+			.optional(),
 		description: z.string().optional(),
 		isTemporary: z.boolean().optional().default(true),
 		visibility: z.nativeEnum(Visibility).default(Visibility.Public).optional(),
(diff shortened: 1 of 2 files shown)


common/models/zod-schemas.ts
```diff
@@ -30,7 +30,11 @@ export const OttApiRequestRoomCreateSchema = z
 			.refine(name => !RESERVED_ROOM_NAMES.includes(name), {
 				message: "not allowed (reserved)",
 			}),
-		title: z.string().max(255, "too long, must be at most 255 characters").optional(),
+		title: z
+			.string()
+			.max(255, "too long, must be at most 255 characters")
+			.refine(s => !/[\n\r]/.test(s), { message: "must not contain newlines" })
```
dyc3, line 36: I would prefer this to not be a regex

common/models/zod-schemas.ts
```diff
@@ -79,7 +83,11 @@ const GrantSchema = z.tuple([z.nativeEnum(Role), z.number()]);
 
 export const RoomSettingsSchema = z
 	.object({
-		title: z.string().max(254).optional(),
+		title: z
+			.string()
+			.max(254)
+			.refine(s => !/[\n\r]/.test(s), { message: "must not contain newlines" })
```
dyc3, line 89: same here

dyc3 (changes requested):

PR #1988, opened 2026-04-03
feat: migrate to drizzle

dyc3:
Migrate from sequelize to drizzle. Why? Sequelize is pretty old, and never really liked it much anyway.

I had the clanker sketch this out to see what exactly this migration would look like. In order to land this, I would have to do a release and version bump.


dyc3:
@pullfrog can you review this again?

PR #2018, opened 2026-05-14
fix: hide Discord login when not configured

kimjune01:
When Discord OAuth credentials are not configured, the login button still appears in the UI and produces an opaque error on click. This change reads the same `DISCORD_CLIENT_ID`/`DISCORD_CLIENT_SECRET` check the auth route already performs and exposes a single helper (`isDiscordLoginEnabled`) consumed by both `user.ts` and `auth/index.ts`, so the UI hides the button when login can't succeed.

The refactor commit factors the duplicated check into `discord-utils.ts` so the two callers can't drift.

## Testing

- `yarn workspace ott-server vitest run tests/unit/auth/discord.spec.ts` (existing tests pass)
- Manual: with no Discord env vars set, login dropdown no longer shows Discord; with valid creds, behavior unchanged.

## Hypothesis graph

Triage notes and reasoning trail: https://github.com/kimjune01/sweep/blob/master/repo-hypotheses/dyc3-opentogethertube.md

The pullfrog bot review surfaced an unauthenticated-button gap the original fix missed; commit 08908d05 closes that case.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -94,6 +94,8 @@ jobs:
       # this is safe because it's only used for testing
       SESSION_SECRET: LERTmi2Jiz6gJHt21AZUWBstezP41P3odxzEsCo1w4zL8XqGyjGdMZ2QPXUenVQQ2fY1xzGBYseB1g9teRHbxF
       CI: true
+      DISCORD_CLIENT_ID: test-id
+      DISCORD_CLIENT_SECRET: test-secret
     services:
       redis:
         image: redis
--- a/client/src/App.vue
+++ b/client/src/App.vue
@@ -222,9 +222,11 @@ const App = defineComponent({
 
 			// ask the server if we are logged in or not, and update the client to reflect that status.
 			const resp = await API.get("/user");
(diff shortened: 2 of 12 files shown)


kimjune01:
Thanks @pullfrog — this was a real gap. The first pass only gated the authenticated account-linking section and the server-side OAuth routes; the unauthenticated `LogInForm` (and the "Link Discord" item in the navbar dropdown) were still rendering the Discord button regardless of whether `DISCORD_CLIENT_ID` / `DISCORD_CLIENT_SECRET` were set. So the button that actually motivated the issue was untouched.

Fixed in 08908d05:
- `GET /api/user` (the unauthenticated response that `App.vue` already hits on every page load) now also returns `discordLoginEnabled`.
- The flag is threaded through the Vuex store and consumed by `LogInForm.vue` and `NavUser.vue`, so the Discord button only renders when OAuth is properly configured.
- Added unauth test coverage in `server/tests/unit/api/user.spec.ts` (both the `discordLoginEnabled=false` default and a `conf.set`-driven `=true` case), and updated the existing strict-equality assertions for the new field.

Server tests pass (37/37) and both server and client typecheck cleanly.

kimjune01:
Cannot validate on current setup, so drafting. Please close or take it over.

PR #2044, opened 2026-06-28
Subtitle rendering via ASS.js for Directly player and custom-manifest

bjuergens:
I added enhanced subtitle rendering via ass.js for custom-manifest videos. 

Changes in this PR:

* include ass.js for rendering ass subtitles via new component "ass-overlay.ts"
* allow multiple vtt and ass subtitle tracks for the same video
* add scrollbars to the overlay-menu for subtitle-selection, because videos can have lots of subtitles. 
* in custom-media-format-jsons  add pseudo mime type "text/x-ass" (ass files don't seem to have an official mime-type associated to them) in addition to text/vtt
* handling of default subtitle track:
  * if there is one, viewers will start viewing with this track enabled. 
  * default subtitle track can be changed while the video is in the queue
  * after it has started, viewers can change the track for themself, and the change won't be synced to other viewers
* i added some line to agents.md, because i was running claude in its webharness and it was running into problems with yarn, which it couldn't recover from. 
* the existing VTT rendering via native browser rendering was left in tact. (it was annoying to integrate both subtitle paths at the same time. I guess architecture would get much simpler if we would remove the native VTT path alltogether and instead render vtt via ass.js as well (i have not tested if ass.js actually can do vtt))

what i have not done yet:

* i have not considered database migrations for existing rooms. for my tests i usually created a clean new room. 
* i did not test against any other video sources, such as youtube. 
* i did not clean up git history (i assume you will just squash this PR)

The actual rendering is isolated to ass-overlay.ts, and seems clean enough to me. The selection of subs seems kinda messy, but i didn't find a way to make it cleaner. I think it is now cleaner that it was before though. Some feedback/guidance would be appreciated. 

I opted for ass.js, as that seemed the to be lightest library that is still being maintained. It seems jassub is more accurate and faster, but also much heaver. The other libraries i found didn't seem to be activly maintained. 

i tested with this deployment: https://ott.wotanii.de/ and this video: https://ottfiles.wotanii.de/public/2026-06-20/clock.mkv.json also i made a small watchparter with some friends, and it worked fine.

during testing i encountered some issues, which are solved now:
* the library ass.js registers a resize-hook to know when to resize the subs. but when i registed the hook before the video was actually loaded, the resizing became weird. I solved it by waiting for the video before intializing the ass-container. 
* claude had trouble running yarn install. I solved it by telling it to use bun instead. I recommend to migrate to bun, as that is in my opinion generally better than npm or yarn. 
* claude had trouble running cargo tests. I solved it by not running cargo tests, and instead deploy to my server for manual testing. clauds issue was that it could not do ipv6.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -16,6 +16,8 @@ OpenTogetherTube is a real-time video synchronization web application. It uses a
 
 ## Build & Development Commands
 
+> **Note:** yarn is prefered. But if that doesn't work in your sandbox, use bun instead.
+
 **Build:**
 
 ```bash
@@ -62,7 +64,7 @@ yarn test:e2e:ui       # Open Playwright UI (interactive)
 yarn test:component    # Run component tests
 
 # Rust tests
-cargo test
+cargo test                # requires ipv6
 cargo test -p ott-balancer
 ```
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -16,8 +16,6 @@ OpenTogetherTube is a real-time video synchronization web application. It uses a
 
 ## Build & Development Commands
 
-> **Note:** yarn is prefered. But if that doesn't work in your sandbox, use bun instead.
-
 **Build:**
 
 ```bash
--- a/client/src/components/VideoQueueItem.vue
+++ b/client/src/components/VideoQueueItem.vue
@@ -186,7 +186,42 @@
 						{{ $t("video-queue-item.edit.title") }}
 					</DialogTitle>
 				</DialogHeader>
-				<Field>
+				<Field v-if="isManifestItem">
+					<FieldLabel for="edit-default-subtitle-track">
+						{{ $t("video-queue-item.edit.default-subtitle-track") }}
+					</FieldLabel>
+					<Spinner v-if="isLoadingManifest" class="size-4" />
+					<FieldDescription v-else-if="manifestError">
+						{{ $t("video-queue-item.edit.manifest-load-failed") }}
+					</FieldDescription>
+					<Select v-else v-model="editedDefaultTrack">
+						<SelectTrigger
+							id="edit-default-subtitle-track"
+							data-cy="edit-default-subtitle-track"
+						>
+							<SelectValue />
+						</SelectTrigger>
+						<SelectContent>
+							<SelectItem :value="TRACK_MANIFEST_DEFAULT">
+								{{ $t("video-queue-item.edit.manifest-default") }}
+							</SelectItem>
+							<SelectItem :value="TRACK_NONE">
+								{{ $t("video-queue-item.edit.no-subtitles") }}
+							</SelectItem>
+							<SelectItem
+								v-for="track in manifestTracks"
+								:key="track.url"
+								:value="track.url"
+							>
+								{{ formatTrackLabel(track) }}
+							</SelectItem>
+						</SelectContent>
+					</Select>
+					<FieldDescription v-if="!manifestError">
+						{{ $t("video-queue-item.edit.default-subtitle-track-hint") }}
+					</FieldDescription>
+				</Field>
+				<Field v-else>
 					<FieldLabel for="edit-subtitle-url">
 						{{ $t("video-queue-item.edit.subtitle-url") }}
 					</FieldLabel>
@@ -233,6 +268,13 @@ import {
 import { Field, FieldDescription, FieldLabel } from "@/components/ui/field";
 import { Icon } from "@/components/ui/icon";
 import { Input } from "@/components/ui/input";
+import {
+	Select,
+	SelectContent,
+	SelectItem,
+	SelectTrigger,
+	SelectValue,
+} from "@/components/ui/select";
 import { Spinner } from "@/components/ui/spinner";
 import { Tooltip, TooltipContent, TooltipTrigger } from "@/components/ui/tooltip";
 import {
@@ -253,6 +295,7 @@ import { API } from "@/common-http";
 import { secondsToTimestamp } from "@/util/timestamp";
 import { ToastStyle } from "@/models/toast";
 import type { QueueItem, VideoAdd } from "ott-common/models/video";
+import type { CustomMediaManifest, CustomMediaTextTrack } from "ott-common/models/zod-schemas";
 import { QueueMode } from "ott-common/models/types";
 import { useStore } from "@/store";
 import toast from "@/util/toast";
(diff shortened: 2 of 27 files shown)


client/src/components/players/DirectPlayer.vue
```diff
@@ -119,71 +134,58 @@ function isCaptionsSupported(): boolean {
...
-		videoMime.value !== "application/json" &&
-		!manifest.value?.textTracks &&
-		!subtitleUrl.value
-	) {
+	assOverlay.destroy();
+}
+
+function applyCaptions(): void {
+	clearRendering();
+	if (!captions.isCaptionsEnabled.value) {
```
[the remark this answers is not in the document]
bjuergens: in an earlier version i did something similar, but it was adding complexity to the dataflow and event flows. I intentionally traded the extra network trip for a simpler implementation.

client/src/components/composables/ass-overlay.ts
```diff
@@ -0,0 +1,124 @@
...
+	// Returns whether the overlay is now active; failures are logged and reported as false, not thrown.
+	async function fetchAndCreate(
+		url: string,
+		video: HTMLVideoElement,
+		box: HTMLElement,
+	): Promise<boolean> {
+		const seq = ++loadSeq;
+		currentUrl = url;
+		try {
+			const response = await fetch(url);
```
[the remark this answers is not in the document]
bjuergens: yes, that is true. I think, that the mp4 files themselves also need cors, so i left it out. 

But cors as annoying and most manifest authors will not be aware of it, so explicit documentation is a good idea.

AGENTS.md
```diff
@@ -16,6 +16,8 @@ OpenTogetherTube is a real-time video synchronization web application. It uses a
 
 ## Build & Development Commands
 
+> **Note:** yarn is prefered. But if that doesn't work in your sandbox, use bun instead.
```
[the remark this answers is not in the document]
bjuergens: The line allows for more different kind of agents to work on the repo. Also I think the project should migrate to bun (but not in this PR)

bjuergens:
i am suprised how well the pullfrog review is. i did some review rounds with opus 4.8, and it never reached the same quality as this pullfrog review. 

I will work on it, but it will take some time.

client/src/components/composables/ass-overlay.ts
```diff
@@ -0,0 +1,143 @@
+import { onBeforeUnmount, type Ref, ref } from "vue";
+import ASS from "assjs";
+import { i18n } from "@/i18n";
+import toast from "@/util/toast";
+import { ToastStyle } from "@/models/toast";
+
+function notifySubtitleLoadFailed(): void {
+	toast.add({
+		style: ToastStyle.Error,
+		content: i18n.global.t("room.subtitle-load-failed"),
```
dyc3: these should be just `i18n.t`, like it is in other files

```suggestion
		content: i18n.t("room.subtitle-load-failed"),
```

AGENTS.md
dyc3, line 1: revert

AGENTS.md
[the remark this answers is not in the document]
dyc3: i am also a bun enjoyer, but there is nothing preventing agents from using yarn

docs/api.yaml
```diff
@@ -686,10 +686,16 @@ components:
         id:
           type: string
           example: dQw4w9WgXcQ
-        subtitleUrl:
+        defaultSubtitleTrack:
```
dyc3: existing persisted room queues are going to be affected by this. remember that this is a live service as well as a self hosted application. 

this is also changing the semantic meaning behind this field. instead of just being a subtitle track its now a "default" one. its a distinction that has no benefit.

common/subtitle.ts
```diff
@@ -0,0 +1,37 @@
+import type { CustomMediaTextTrack } from "./models/zod-schemas.js";
+
+export function normalizeSubtitleTrack(value: string | null | undefined): string | null {
+	return value || null;
+}
+
+function subtitleUrlExtension(url: string): string | undefined {
+	const path = url.split("?")[0].split("#")[0];
+	return path.split(".").pop()?.toLowerCase();
+}
```
dyc3, line 14: you shouldn't be parsing urls by hand, there are existing tools for this

common/subtitle.ts
```diff
@@ -0,0 +1,37 @@
...
+		return "text/vtt";
+	}
+	return null;
+}
+
+export function externalSubtitleAsTextTrack(url: string): CustomMediaTextTrack {
+	const contentType = inferSubtitleContentTypeOrNull(url);
+	if (!contentType) {
+		// Callers only reach here with a server-validated url, so this is a programming error.
+		throw new Error(`Cannot build a text track for unsupported subtitle url: ${url}`);
```
dyc3: this is called by a `computed`. if it throws, the failure mode is to not render ui rather aggressively, which could be bad.

dyc3 (changes requested):
Sorry this took so long to review.

docs/api.yaml
bjuergens: yes, i agree. i am reverting that thing.

bjuergens:
> Sorry this took so long to review.

haha, same here. I had to deal with some irl issues and didn't find the time to finalize my refactorings of the previous round of reviews (most importantly the subtitletrack->defaultsubtitletrack thing)

bjuergens:
I remember why I did the defaultSubtitleTrack thing. It was because there are multiple relevant subtitles for each file:

* the list of subtitles from the custom manifest
* the explicit default subtitle from the custom manifest
* the implicit default subtitle (IIRC the first sub, if there is no default)
* the overwrite default subtitle, which a user can apply while the video is in the queue
* the initial subtitle track, which is selected when the video starts playing
* the actual subtitle track, which is currently selected in the player

I decided that the default track shall be managed by the room and will be applied at the start of a video for each viewer, and the currently playing subtitle is fully controlled by each player individually and will not be synced between the peers. To make that distinction clear, I renamed the schema field, because I (and also Claude) kept getting confused there.

And since I was working under the assumption that only my deployment would be affected, I just renamed it. Only later did I consider opening a PR, and I didn't consider that longer-running instances may need to be migrated.

client/src/components/players/DirectPlayer.vue
```diff
@@ -16,51 +16,66 @@
...
+	if (videoMime.value === "application/json") {
+		return manifest.value?.textTracks ?? [];
+	}
+	if (defaultSubtitleTrack.value) {
+		return [externalSubtitleAsTextTrack(defaultSubtitleTrack.value)];
+	}
+	return [];
+});
+const vttTracks = computed(() =>
+	textTracks.value.filter(track => track.contentType === "text/vtt")
```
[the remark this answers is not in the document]
bjuergens: i added the supression. 

I considered extracting it into a separate function, or changing it to a for loop. both would fix the lint bug, but they wouldn't even improve the code, i think. And with the surpression, it's easier to revisit in case biome solved their bug here.

AGENTS.md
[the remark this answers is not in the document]
bjuergens: reverted

client/src/components/composables/ass-overlay.ts
bjuergens: `i18n.t()` didnt seem to work. i went with 

```
const t=useI18n()
...
content: t("room.subtitle-load-failed"),
```

docs/api.yaml
bjuergens: reverted the server-side thing. but I kept the distinction in the frontend-side. See the separat toplevel comment

server/tests/unit/api/room.spec.ts
```diff
@@ -483,7 +483,7 @@ describe("Room API", () => {
 				.send({
 					service: "direct",
 					id: "foo",
-					subtitleUrl: "https://example.com/subtitles.vtt",
```
dyc3, line 486: nit: revert - this is changing an existing test to assert new behavior

server/tests/unit/api/room.spec.ts
```diff
@@ -496,13 +496,39 @@ describe("Room API", () => {
 					expect.objectContaining({
 						service: "direct",
 						id: "foo",
-						subtitleUrl: "https://example.com/subtitles.vtt",
```
dyc3, line 499: nit: revert

client/src/components/VideoQueueItem.vue
```diff
@@ -186,13 +186,42 @@
...
+								:key="track.url"
+								:value="track.url"
+							>
+								{{ formatTrackLabel(track) }}
+							</SelectItem>
+						</SelectContent>
+					</Select>
+					<FieldDescription>
+						{{ $t("video-queue-item.edit.default-subtitle-track-hint") }}
+					</FieldDescription>
```
dyc3, line 215: the entire concept of being able to set a "default subtitle track" is a completely different feature than just .ass subtitle support. please remove it.

server/services/direct.ts
```diff
@@ -134,6 +134,8 @@ export default class DirectVideoAdapter extends ServiceAdapter {
 			length: Math.ceil(manifest.duration),
 			thumbnail: manifest.thumbnail,
 			mime: "application/json",
+			textTracks: manifest.textTracks,
+			subtitleUrl: manifest.textTracks?.find(t => t.default)?.url ?? null,
```
dyc3, line 138:
```suggestion
```

server/tests/unit/room.spec.ts
```diff
@@ -263,7 +263,7 @@ describe("Room", () => {
 			});
 
 			it("should preserve subtitleUrl from PlayNowRequest", async () => {
-				const subtitleUrl = "https://example.com/subtitles.vtt";
+				const subtitleUrl = "https://example.com/track.de.ass";
```
dyc3, line 266: nit: revert

client/src/components/controls/VideoSettings.vue
```diff
@@ -60,26 +60,28 @@
...
-						v-for="(quality, idx) in qualities.videoTracks.value"
-						:key="idx"
-						type="button"
-						class="menu-item"
-						:class="{ 'menu-item-active': idx === qualities.currentVideoTrack.value }"
-						@click="selectQuality(idx)"
-					>
-						{{ formatQuality(quality) }}
-					</button>
+					<div class="menu-scroll">
```
dyc3, line 63: these `menu-scroll` additions are unrelated - remove them

dyc3 (changes requested):
There's a lot of changes that are completely unrelated to adding .ass subtitle support. This makes it really hard to review.

bjuergens:
> scope

I understand the need to have focussed PR to make it reviewable/mergeable. Especially, since my changes grew organically (i.e. I was scratching my own itch, and was just doing lots of things, that I wanted/needed for my own usecase namely "anime watch party"), and the PR only came as an after thought. 

I think the cleanest solution would be:
1. I address all the comments you made so far on this branch/PR
2. I turn that branch into the basebranch of my fork
3. I create a new branch for the PR (or separate PR), which only contains the minimal set of changes needed to get ASS subs working, based on upstream's master
4. After that has been merged upstream, I rebase my fork's basebranch on there
5. Optional followup: I create issues and/or PR for other things (e.g. adding scrollbars to subtitle selection for that case that there are more than 20)

That would be clean, i think, but it would also be a lot of work. So much in fact, that I am not sure if it will ever be done. I guess you could just wait a month or two to see if I made any progress, before you make a decision how to go forward.

PR #2054, opened 2026-07-22
feat(client): add an opt-in setting for YouTube's native control bar

WuTofu:
Adds an opt-in client setting to show youtube's native control bar instead of ott's own controls, while keeping ott authoritative over anything that must stay room-synced.

- New "Use youtube's native control bar" setting and it's off by default.
- `YoutubePlayer` accepts a `nativeControls` prop: enables the iFrame controls, and the native fullscreen button (`fs: 0`) to avoid double-handling with ott's own shortcuts/fullscreen logic.
- Volume/mute are the only native inputs forwarded bidirectionally: since the iframe api has no change events for them, they're polled every 250ms and diffed against last-observed values (`pollNativeVolume`), then surfaced as `user-volume-change` / `user-mute-change` for `OmniPlayer` to fold into the shared `useVolume()` state.
- Playback rate stays ott-controlled: native rate changes are reported (`native-rate-change`) but snapped back to the room's rate in `OmniPlayer.onNativeRateChange`.


WuTofu:
@pullfrog review again

(Edit: sigh, don't know how to interact with this bot correctly)

dyc3:
It's not you, its set up to only listen to me

dyc3:
@pullfrog can you review this again

dyc3:
For context, I've considered doing this before, but abstained because I thought it would be too complicated.

WuTofu:
> For context, I've considered doing this before, but abstained because I thought it would be too complicated.

I'm okay with closing this pr if you still feel this way😅

dyc3 (changes requested):
This would be nice to have, but not as the default. It would let users set video quality manually for youtube videos. That's something that's been sorely missing, particularly because youtube took away the video quality api. But, its still an extremely risky change. A rather fundamental architecture decision I made years ago was that OTT takes over the responsibility of playback, and all controls must route through OTT's logic and user interface, to intercept all of that.

I think this implementation is brittle because the event echo detection is purely time based. I think there's a better solution.

This is also missing handling of `onPlaybackRateChange`. What happens when a user changes the playback speed in the youtube native controls? Also, what about fullscreen mode? Keyboard controls don't work once you've focused the iframe either... 

The biggest benefit of this is being able to set video quality (unless there's something else im missing, feel free to elaborate). So I would rather this be a manual toggle of some kind, probably a client setting, that's off by default, and it simply enables the native controls without any guarantee that they actually work.

WuTofu:
> I think this implementation is brittle because the event echo detection is purely time based. I think there's a better solution.

The more I think about this, the more I get why you said it would be too complicated.
So, I am thinking if we should just override the youtube player's status (play/pause/time/playbackRate) when youtube player fires `onStateChange` or `onPlaybackRateChange` and show a pop-up to remind users to use OTT's control only.

@dyc3 does that work to you?

dyc3:
> override the youtube player's status

this is already how it works, at least for play/pause. so yes, that makes the most sense.

I don't think there should be a pop up when the state changes. It should be clear when native controls are toggled on. This can just be a note in the client settings.

dyc3:
When testing this locally, it takes a really long time to finish joining the room for some reason. I'll have to look at it later.

WuTofu:
@dyc3 could you show me how to reproduce the issue you encountered?

dyc3 (approved):
I think this is about as good as its going to get.

I can't reproduce the joining bug anymore. I suspect it was just my browser being slow.

PR #2067, opened 2026-08-22
feat: add native Jellyfin support

stephenvaz:
## What

Adds native Jellyfin video service support to OpenTogetherTube, allowing users to paste Jellyfin URLs directly into the room queue.

## Why

Previously, users had to manually construct HLS URLs and subtitle links. This provides a seamless experience similar to YouTube/Vimeo integration.

### Server
- **New `jellyfin.ts` adapter**: Parses Jellyfin web UI, download, and HLS URLs. Extracts API keys from query params or hash fragments. Uses compound IDs (`serverOrigin::itemId`) for server identification.
- **Auth**: Sends X-Emby-* headers and X-Emby-Token for Jellyfin 10.11+ API compatibility. Caches user IDs per server+apiKey.
- **Video info**: Fetches item details via `/Users/{userId}/Items/{itemId}`, constructs HLS stream URLs, and extracts subtitles from PlaybackInfo API.
- **Series resolution**: Resolves series/season URLs to individual episode videos.
- **Config**: Added `jellyfin.instances` for server allowlist (empty = allow all).

### Client
- **HlsPlayer**: Rewrote caption handling to use native `<track>` elements instead of HLS.js subtitle tracks, enabling external subtitle injection. Added Jellyfin bitrate profiles and URL-based quality switching.
- **Quality**: Auto quality label now displays "Auto (X Mbps)" using the current MaxStreamingBitrate from the URL.
- **CC button**: Fixed to select track 0 on first press if no track was previously selected, remembering the last selection on subsequent toggles.
- **VideoSettings**: Added scrollable subtitle menu for long lists.

### Tests
- 31 unit tests covering URL parsing, API key extraction, video info fetching, series resolution, and auth headers.


dyc3:
@pullfrog review it again, and resolve any fixed issues from your previous review.

stephenvaz:
There are a few bugs with the Jellyfin bitrate change implementation, once those are ironed out the review can be started

client/src/components/controls/ClosedCaptionsSwitcher.vue
```diff
@@ -38,6 +38,11 @@ const supported = computed(() => {
 const enabled = computed(() => captions.isCaptionsEnabled.value);
 
 function toggleCaptions() {
+	if (!captions.isCaptionsEnabled.value) {
+		if (captions.currentTrack.value === null && captions.captionsTracks.value.length > 0) {
+			captions.currentTrack.value = 0;
+		}
+	}
```
dyc3, line 45: This seems unrelated

client/src/components/controls/VideoSettings.vue
dyc3, line 1: this too

client/src/components/players/HlsPlayer.vue
```diff
@@ -30,19 +30,34 @@ import type {
...
+});
+const { videoUrl, thumbnail, service } = toRefs(props);
 const videoElem = ref<HTMLVideoElement | undefined>();
 const captions = useCaptions();
 const qualities = useQualities();
 const audioBoost = useMediaAudioBoost(videoElem);
+const currentUrl = ref(props.videoUrl);
 let hls: Hls | undefined;
 
+const isJellyfin = computed(() => service.value === "jellyfin");
```
dyc3, line 59: why does the hls player need to care about jellyfin if its just HLS streams?

server/services/jellyfin.ts
dyc3, line 1: This is lacking sufficient documentation and references on how jellyfin works. I can't properly assess your implementation without that.

dyc3 (changes requested):

client/src/components/controls/ClosedCaptionsSwitcher.vue
stephenvaz: From a UX point of view, if a user taps on the subtitle button and it goes to the enabled state then it should pick the 0th track instead of going into an incorrect state (as of now)

client/src/components/players/HlsPlayer.vue
stephenvaz: For files where it's not re-encoded into different bitrates at the time of storing, jellyfin web player provides step up options (720kbps, 1.5mpbs, 3mpbs, 6mbps, 10mbps). When a user selects a different bit rate, it is transcoded and streamed in realtime but for this you need a new PlaybackSessionId that's generated at the jellyfin server, the old one can't be re-used which requires the url to also be updated. 

Since this is tied to jellyfin specifically we need to have an understanding of how to handle quality changes. 

But i'm a bit inclined to scrap the quality part for now, it's a bit of a hacky solution (hence moved the MR to draft)

server/services/jellyfin.ts
stephenvaz: Sure, i'll add documentation for this and move it to out of the draft state when i feel it's ready for review

client/src/components/controls/ClosedCaptionsSwitcher.vue
dyc3: The "fix" is still completely unrelated to adding jellyfin support.

client/src/components/players/HlsPlayer.vue
dyc3: Hm, yeah that kinda sucks to deal with. Honestly, I would be leaning towards having a new player then. I would prefer the generic players to stay as generic as possible.

